PR #31, opened 2016-08-30
Implement the requests to work with the (WP)REST-API

rensw90:
This PR fixes a couple of things so that the wizard works correctly when loaded as a dependency in wordpress-seo. 

This PR fixes the following things:
- When saving the options, a PUT requests is send instead of a POST request.
- The JSON enconding is now correct, previously this was double encoded.
- The PUT requests now contain headers which are needed to put the WP nonce in the headers.
- The componentName in the config is renamed to the old component, because the WP-API expects this.

Also incorrect default props in the OnboardingWizard are fixed.

Testing:
1. Run grunt build and the npm start.
2. Go through the steps and see if everything goes ok and no errors are shown in the console.
3. Check if a PUT request is send by the wizard and the params for the request are JSON.
4. By default jQuery is not present, to test the jQuery functionalities add ```import jQuery from "jquery";``` to the top of postJSON.js. If jquery is not yet installed run: ``npm install jquery``
5. Now check if a put request is send when you go to the next or previous step. This request has to contain: ```"X-WP-Nonce": "test1234"``` in the headers.


App.js
```diff
@@ -35,6 +35,14 @@ class App extends React.Component {
 			case "wizard":
 			default:
 				let config = cloneDeep( Config );
+				let endpoint = {
+					url: "http://127.0.0.1:8882/onboarding",
+					headers: {
+						"X-WP-Nonce": "test1234",
```
diedexx: please change the nonce to something more generic

App.js
```diff
@@ -35,6 +35,14 @@ class App extends React.Component {
 			case "wizard":
 			default:
 				let config = cloneDeep( Config );
+				let endpoint = {
+					url: "http://127.0.0.1:8882/onboarding",
```
diedexx: I don't think this should be localhost.

App.js
```diff
@@ -35,6 +35,14 @@ class App extends React.Component {
 			case "wizard":
 			default:
 				let config = cloneDeep( Config );
+				let endpoint = {
```
diedexx: The naming seems a bit off.. an endpoint doesn't have headers and (correct me if i'm wrong) an endpoint _is represented_ by a url.

composites/OnboardingWizard/OnboardingWizard.js
```diff
@@ -271,15 +262,13 @@ class OnboardingWizard extends React.Component {
 }
 
 OnboardingWizard.propTypes = {
-	endpoint: React.PropTypes.string.isRequired,
+	endpoint: React.PropTypes.object.isRequired,
 	steps: React.PropTypes.object.isRequired,
+	fields: React.PropTypes.object.isRequired,
 	currentStepId: React.PropTypes.string,
```
diedexx: This prop appears to be unused.

composites/OnboardingWizard/config/production-config.js
```diff
@@ -11,25 +11,25 @@ let configuration = {
 	},
 	"fields": {
 		"upsellConfigurationService": {
-			"componentName": "HTML",
+			"component": "HTML",
```
diedexx: I disagree with changing the name of this property. What is the reason to change it to `component`?

composites/OnboardingWizard/helpers/postJSON.js
```diff
@@ -46,46 +41,60 @@ let postJSONFetch = ( url, data ) => {
...
  * Wrapper method when fetch should be used.
  *
  * @param {string} url  The endpoint to send the data to.
+ * @param {object} headers Object containing the headers for the request.
  * @param {Object} data The JSON object to send to the server.
  *
  * @returns {Promise} Returns a wrapped promise.
  */
-let postJSON = ( url, data = {} ) => {
+let postJSON = ( url, headers = {}, data = {} ) => {
```
diedexx: I personally would put headers as the third parameter, since you will most likely pass your own data to a POST/PUT request, but are less likely to add extra headers. This prevents the user from having to add an unnecessary parameter.

composites/OnboardingWizard/helpers/postJSON.js
```diff
@@ -46,46 +41,60 @@ let postJSONFetch = ( url, data ) => {
...
-			      ( response ) => {
-				      resolve( response );
-			      }
-		      )
-		      .fail(
-			      () => {
-				      reject( "Wrong request" );
-			      }
-		      );
+let postJSONjQuery = ( url, headers, data ) => {
```
diedexx: Since you use PUT requests instead of POST request, you should rename some variables in this file and also the filename.

index.html
```diff
@@ -3,7 +3,7 @@
 <head>
     <meta charset="UTF-8">
     <title>Title</title>
-    <link rel="stylesheet" href="dist/yoast-components.min.css" />
+    <link rel="stylesheet" href="dist/yoast-components-standalone.min.css" />
 </head>
 <body style="background-color: #f0f0f0;">
```
diedexx: Should this not be in the scss file?

diedexx:
CR Done 👷

composites/OnboardingWizard/config/production-config.js
rensw90: Yup, the wordpress API already uses `component`. If this is changed the wizard does not work in WP anymore.

rensw90:
Also added a function that reads the data values from the config if the fieldValues are not yet set by the user.

composites/OnboardingWizard/config/production-config.js
diedexx: Can we update the api instead? If we want to rename this,_now_ is the time

PR #40, opened 2016-08-31
Improve the check for non-existing field components

rensw90:
The wizard kept loading when a component did not exist. To prevent this, the check if components exist is improved.

The wizard should now just go to the next step and not render the non-existing component.

Testing:
1. Set a component in the production-config.js to a non existing one.
2. Start the wizard and go to the step with the non existing component.
3. The step should just render.


composites/OnboardingWizard/Step.js
```diff
@@ -38,6 +38,7 @@ class Step extends React.Component {
 	 * @returns {void}
 	 */
 	componentWillReceiveProps( props ) {
+		this.key
```
andizer: What is this?

composites/OnboardingWizard/Step.js
```diff
@@ -104,8 +105,10 @@ class Step extends React.Component {
 		return keys.map( ( name, key ) => {
 			let currentField = fields[ name ];
 
-			if ( Components[ currentField.componentName ] === "undefined" ) {
-				return;
+			if ( typeof Components[ currentField.componentName ] === "undefined"
+			     | !Components[ currentField.componentName ] ) {
```
andizer: Why one `|`

PR #44, opened 2016-09-02
Add html decoder for separator entities

rensw90:
Separators are stored as html entities in the config. For this a HTML decoder is added. This translates the entities back to their symbols.

### DIFF
--- a/composites/OnboardingWizard/helpers/htmlDecoder.js
+++ b/composites/OnboardingWizard/helpers/htmlDecoder.js
@@ -1,3 +1,9 @@
+/**
+ * @summary Converts HTML entities to the actual symbols.
+ *
+ * @param htmlText The HTML text.
+ * @returns {string} The string with converted HTML entities.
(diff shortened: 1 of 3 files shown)


composites/OnboardingWizard/helpers/htmlDecoder.js
```diff
@@ -0,0 +1,7 @@
+let decodeHTML = ( htmlText ) => {
```
andizer: Missing docs.

composites/OnboardingWizard/helpers/htmlDecoder.js
```diff
@@ -0,0 +1,15 @@
+/**
+ * @summary Converts HTML entities to the actual symbols.
+ *
+ * Source: http://stackoverflow.com/questions/3700326/decode-amp-back-to-in-javascript
```
andizer: Awesome

PR #50, opened 2016-09-07
Add optimizations needed for the implementation of custom components

rensw90:
- Updated the postJSON helper to a more generic ajaxHelper.
- Added some checks needed for the MailChimpSignup custom component.

Testing:
1. Start the wizard.
2. Go through the steps and see if the data is correctly send. You can check this in the params in the response. The requests should be PUT requests.

### DIFF
--- a/composites/OnboardingWizard/Step.js
+++ b/composites/OnboardingWizard/Step.js
@@ -109,13 +109,24 @@ class Step extends React.Component {
 		let keys = Object.keys( fields );
 
 		return keys.map( ( name, key ) => {
+
 			let currentField = fields[ name ];
 
 			if ( typeof this.components[ currentField.componentName ] === "undefined"
 			     || ! this.components[ currentField.componentName ] ) {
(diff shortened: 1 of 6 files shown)


composites/OnboardingWizard/Step.js
```diff
@@ -109,13 +111,24 @@ class Step extends React.Component {
...
 			     || ! this.components[ currentField.componentName ] ) {
-				console.error(`Trying to load non-existing component: ${currentField.componentName}`);
+				console.error( `Trying to load non-existing component: ${currentField.componentName}` );
 				return null;
 			}
+
+			/* The last Mailchimp signup does not have to be rendered,
+			 * if it is completed in the first signup component.
+			 */
+			if ( currentField.componentName === "MailchimpSignup" && this.props.currentStep === "success" ) {
```
andizer: Since the mailchimp signup is a custom component, this shouldn't be done here.

PR #53, opened 2016-09-09
Because of the old value came back.

### LINKED ISSUE #52: When I empty a field with a predefined value the predefined value is re-inserted
When I empty a field in the wizard the original value is re-inserted into the field.

andizer:
The old value still returns when entering a new value.

Fixes #52


rensw90:
Cr + Acceptance done 👍

PR #59, opened 2016-09-13
Make the conditional fields generic.

andizer:
Makes the conditional fields generic and allows each element to have a conditional config to rely on.


terw-dan:
CR: you are removing some tests (since there is no conditional field), but I don't see a test added for the hasFieldValue, filterConditionalFields en showconditionalFields functions.

terw-dan:
CR :+1:

terw-dan:
Acceptance :+1:

PR #62, opened 2016-09-13
Fix keys not unique for input components.

rensw90:
The input fields did not get unique keys in React. This caused some fields to get the value from another field when a step is re-rendered.

This also fixes the problem with the radio buttons from react 15.3.1, so the React version in the package is also set to 15.3.1.

Testing:
1. Start the wizard and go to step 4.
2. Fill in some text in the textfield.
3. Go to the next step and see if no input fields in that step have the value from the previous step.

Also test the radio buttons.

### DIFF
--- a/composites/OnboardingWizard/Step.js
+++ b/composites/OnboardingWizard/Step.js
@@ -138,7 +138,8 @@ class Step extends React.Component {
 				return null;
 			}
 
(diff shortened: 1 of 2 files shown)


atimmer:
CR Done 👍🏻

PR #64, opened 2016-09-13
Localize the onboarding wizard

atimmer:
This makes it possible to translate the next and previous buttons

Inspired by i18n-calypso, but https://github.com/Automattic/i18n-calypso/issues/19 blocks us from using it.

Required for https://github.com/Yoast/wordpress-seo/pull/5582


utils/i18n.js
```diff
@@ -0,0 +1,81 @@
...
+	}
+}
+
+/**
+ * A higher order component to add a translate function.
+ *
+ * @param {Object} ComposedComponent The original React component.
+ */
+export function localize( ComposedComponent ) {
+	let componentName = ComposedComponent.displayName || ComposedComponent.name || '';
```
jcomack: Please use double quotes.

utils/i18n.js
```diff
@@ -0,0 +1,81 @@
...
+	let i18nProps = {
+		translate: translate,
+	};
+
+	return React.createClass( {
+		displayName: 'Localized' + componentName,
```
jcomack: See previous comment.

jcomack:
CR done. 👷

Acceptance done as well. Translations work, however I do get the following error.

yoast_seo_ _setup_wizard

I cannot get rid of this issue despite clearing out my `node_modules` and reinstalling. Also, for some reason, I cannot see why the Travis builds keep failing, as Travis says the repository doesn't exist.

atimmer:
The error has nothing to do with this PR. I have fixed it on `develop` when not using `npm link`.

PR #70, opened 2016-09-14
Onboarding Wizard initial focus proof of concept

### LINKED ISSUE #69: Focus on Next and Previous buttons
See https://github.com/Yoast/wordpress-seo/issues/5528
On the Onboarding Wizard, when using only the keyboard and pressing the "Next" and "Previous" buttons, focus stays on the buttons. This is not ideal for accessibility for the reasons mentioned in https://github.com/Yoast/wordpress-seo/issues/5528

It seems to me the buttons are not completely re-rendered on each step, while their properties yes, they get updated. This is good for a11y because screen readers may get confused then elements are re-rendered. On the other hand, the buttons have focus and even though the step is updated, the focus stays on the button. This is native browsers behaviour :)

Ideally, when pressing "Next" and "Previous" the focus should be set  on a proper element at the top of the wizard.

afercia:
This is a proof of concept. Experimenting a way to set the initial focus on the active step when pressing the "Next" and "Previous" buttons. Things can be coded better for sure, needs some initial feedback though.

For now, focus is moved to the active step button in the step indicator. Not ideal because then users will heave to tab through all the step buttons but at least they won't skip any important content. Suggestions for improvements welcome.

For testing purposes, added a strong focus style to help while testing, this should be eventually removed of course. Build the css to see what happens when pressing the next/previous buttons:

screen shot 2016-09-14 at 15 56 07

Fixes #69


afercia:
Fixes #69 
Connects to #69

atimmer:
@afercia Let's put the focus on the actual form because that is what changed. My initial thought is to put the focus on the heading element. An alternative would be adding a `screen-reader-text` element with `Step 4: Company or person` which will be focused after going to a certain step.

I think there should also be a `screen-reader-text` on the next button telling the user that the focus will be on the form in the next step. The same goes for the previous button.

afercia:
> Let's put the focus on the actual form
Yep, that would be better :) or we can try to set focus on the step container making it focusable with tabindex=0 and using aria-label. Will try both, either ways yep I agree the best thing is, after pressing the buttons, users land on something that gets announced `Step 4: Company or person`.

afercia:
Focus now is set to the step container, and using `tabindex="-1"` and `aria-labelledby` we can make screen readers announce the title:

screen shot 2016-09-15 at 20 04 52

Added also `aria-label`s on the next/previous buttons.
Made the label for the last button ("Close") translatable.

Todo: 
focus gets moved also when clicking on the step buttons... this should be avoided, any suggestions about the best way to accomplish this are very welcome.

composites/OnboardingWizard/OnboardingWizard.js
```diff
@@ -210,7 +214,8 @@ class OnboardingWizard extends React.Component {
 		let hideButton = false;
 
 		if ( type === "next" && ! currentStep.next ) {
-			attributes.label = "Close";
+			attributes.label = this.props.translate( "Close" );
+			attributes['aria-label'] = this.props.translate( "Close the Wizard" );
 			attributes.onClick = () => {
```
afercia: Not sure about this pre-existing line. `attributes` is an object with several properties, not sure how attaching an `onClick` event is right.

atimmer: This is fairly standard for React components. Instead of thinking as HTML `attributes`, think of them as React component `attributes`.

afercia:
Implemented a way to avoid setting focus when clicking on the step buttons on the top.
Focus is now moved to the content just when clicking next/previous. Had to find a way to get _which_ element was clicked to distinguish the two cases. Open to suggestions for a better way.

CSS:
- focus style on the main content is reset to `outline: none`
- improved the `;hover` style on the Next button.

Todo:
- implement a `:focus` style for the next/previous buttons but maybe this can also be done in a separate issue with a more general solution for form elements.

afercia:
Added `:focus` style for the Next/Previous buttons and solved conflicts.

terw-dan:
CR and Acceptance :+1:

PR #75, opened 2016-09-15
Set finish url to go to when wizard has been finished

andizer:
Use a finish url in the config. When pressing the close button in the wizard, this url is used to navigate to.


terw-dan:
CR  :+1:

terw-dan:
Acceptance :+1:

PR #80, opened 2016-09-16
Set vertical align middle on the active step button

### LINKED ISSUE #76: Alignment of navigation bullets is of in some browsers.
When the standalone onboard wizard is running in a very small viewport:
image
The colored bullet has a strange alignment. Also the texts are missing (is this intended?). 
image
Also on wider viewports.

This is in FireFox in Ubuntu. 

Chromium in Ubuntu has it's alignment better, and visible text in it's smallest viewports. 
image

But the grey line is running through the text.

afercia:
Fixes #76 

Should look better now. Needs to be tested on Firefox.


afercia:
Todo:
on smaller viewports, the buttons label is visible on Chrome (and the grey line is running through the text) but hidden on Firefox. I guess it should be always hidden?

screen shot 2016-09-16 at 09 27 09
screen shot 2016-09-16 at 09 27 01

afercia:
Re: buttons labels
looks like there's a different implementation across browsers in the way they support `text-indent` on flex items and this is one of the reasons we should be super careful when using flexbox :). Hiding the text seems intentional since in the CSS there's `text-indent: -9999px;` but this doesn't work on Chrome. Looking for some solution.

afercia:
The responsive view in Safari seems even more tricky (don't know how much the Safari emulator is reliable though). Incoming fix, but this should be tested on real iOS devices.

afercia:
Fixed in Chrome, Firefox, Safari on OS X. The buttons text is now hidden on all these browsers, and the clickable area is a bit reduced but still usable (see below).

Needs to be tested on Windows and real iOS devices.

screen shot 2016-09-16 at 11 02 52

atimmer:
CR Done 👍🏻

terw-dan:
Tested this on Ubuntu in Firefox and Chromium. There it looks fine.

andizer:
I think the target branch of this pull should be `release/1.0`

afercia:
Changed target branch to `release/1.0` after rebasing.

atimmer:
So this has been tested and code reviewed: Merging.

PR #81, opened 2016-09-16
Add screen reader text for choice component

rensw90:
## Summary
The title separators are choice components. The special thing about this is that they contain a symbol and not text for a choice. The screen reader does not always give a good description for these symbols.

A solution for this is needed.

This PR can be summarized in the following change-log entry:
- Add more descriptive screen reader texts for choosing title separators.

Normally a label is sufficient for a screen reader, but now the labels contain symbols. The symbols are not always pronounced correctly. To override the default(visible) label content an aria-label is added. This replaces the symbol that the label contains with a more descriptive text for the screen reader. 
*

## Test instructions

- Go to step 9 'title separators' in the wizard.
- Inspect the separators and see if the labels have an 'aria-label' property.
- Enable the screen reader (CMD+F5) and see(or actually hear) if the screen reader reads the aria label instead of the actual label, it works correctly. You can test this with the vertical bar, by default 'vertical line' is read, but when the aria-label works 'vertical bar' is read.

*
Fixes https://github.com/Yoast/wordpress-seo/issues/5608

### DIFF
--- a/composites/OnboardingWizard/components/Choice.js
+++ b/composites/OnboardingWizard/components/Choice.js
@@ -30,7 +30,7 @@ const Choice = ( props ) => {
 						<Input name={fieldName} type="radio" label={choice.label} onChange={props.onChange}
 						       value={choiceName} optionalAttributes={{ id, checked }}
 						/>
-						<Label for={id}>{htmlDecoder(choice.label)}</Label>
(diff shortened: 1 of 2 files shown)


atimmer:
CR Done 👍🏻

terw-dan:
Acceptance :+1:

PR #86, opened 2016-09-19
Use pointer-events: none on the inactive step buttons.

### LINKED ISSUE #85: Safari and click events on svg elements bug
See https://github.com/Yoast/wordpress-seo/issues/5531

Safari doesn't fire click events on svg elements. Researching a bit, the most common solution is setting the CSS property `pointer-events: none;` to the svg element OR one of its ancestors.

afercia:
Fixes #85 

Looks like setting `pointer-events: none` on the svg element is not enough, had to set it on all the button children:
- set `pointer-events: none` on all the button children
- do this only for the inactive step buttons (custom component)

Needs to be tested on iOS devices and IE, to be sure there are no unwanted side effects.


terw-dan:
CR done :+1:

terw-dan:
Acceptance: IE doesn't work due to another bug: https://github.com/Yoast/wordpress-seo/issues/5660

atimmer:
Acceptance done 👍🏻

Fixes it for me on Safari, iOS is unchanged.

PR #91, opened 2016-09-21
Adds babel polyfill for IE compatbility

terw-dan:
Fixes https://github.com/Yoast/wordpress-seo/issues/5660
Requires: https://github.com/Yoast/wordpress-seo/pull/5686
Since Internet Explorer doesn't support `assign` and `promise` we use babel-polyfill so we can use this in Internet Explorer. 
IE 9, 10 and 11 are now supported. 

For testing: Load the onboard wizard in IE. I've checked this in IE11 with the compatibility modes of IE9 and IE10. IE8< are not supported.


rensw90:
CR + Acceptance done 👍

terw-dan:
Moved the polyfill to the application (in this case the main.js). 
For WordPressSEO it is now included in the js/src/configuration-wizard.js

PR #93, opened 2016-09-21
CSS adjustments for the Onboarding Wizard

### LINKED ISSUE #92: Onboarding Wizard CSS adjustments
See also https://github.com/Yoast/wordpress-seo/pull/5681

The Onboarding Wizard uses its own page and doesn't inherit some of the WordPress styles, including the WP font stack. 

Form controls need some slight CSS adjustments mostly related to alignments and spacing.

Links (normal links within blocks of text) need some styling.

Worth mentioning the "native fonts" WordPress font stack will make the fonts look different on different platforms, this affects also form elements, e.g. text input fields.

Adding the a11y label since links styling is important for accessibility.

afercia:
Fixes #92 

See also https://github.com/Yoast/wordpress-seo/pull/5681

- adjustments for form controls alignments
- adds the WordPress font stack
- adds the WordPress styling for links

Note: 
the new font stack will make the fonts look different across different platforms. For example, on OS X the font is now slightly bigger. This will be consistent with what WordPress does, but that's not to say we should like it.

Since the Onboarding Wizard is a "one-page standalone" tool, we can also use a complete differet styling. Feedback welcome.


terw-dan:
CR Done :+1:

terw-dan:
Acceptance: Looks good in Chrome and Firefox in Windows and Chromium and Firefox in Ubuntu.

PR #95, opened 2016-09-22
Convert html entities for choice and text area input fields

rensw90:
## Summary

The labels for Choice input fields like radio buttons did not decode the HTML entities back to their original symbol.

- Decoded the strings for the labels with the htmlDecoder helper.
- The generic component TextArea that is used for Input type fields in the wizard only decodes the HTML if a string is passed as children. Not if JSX.Elements are passed as children.

## Test instructions
1. Add a HTML entity to the label of a radio field. This can be done in the getFieldProps function in the Step component. The label for a Choice element is defined in the props.properties.label variable.
2. Check if the symbol is displayed as the label for the choice component and not the HTML entity.

Fixes https://github.com/Yoast/wordpress-seo/issues/5613

### DIFF
--- a/composites/OnboardingWizard/components/Choice.js
+++ b/composites/OnboardingWizard/components/Choice.js
@@ -39,7 +39,7 @@ const Choice = ( props ) => {
 
 	return (
 		<div className={props.className}>
-			<p className="yoast-wizard-field-description">{props.properties.label}</p>
+			<p className="yoast-wizard-field-description">{htmlDecoder(props.properties.label)}</p>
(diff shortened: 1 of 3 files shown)


terw-dan:
CR :+1:

atimmer:
This would mean that any string that is passed as a label to the Choice component is not escaped, which is essentially the same as using `dangerouslySetInnerHTML`. I don't think this is the right approach.

rensw90:
Replaced with https://github.com/Yoast/wordpress-seo/pull/5700

PR #97, opened 2016-09-22
CSS selector to target all the input text fields.

### LINKED ISSUE #96: Mailchimp input fields width in the onboarding wizard
In the onboarding wizard, all the text input fields have a default (responsive) width of 450 pixels, except the `mailchimpName` and `mailchimpEmail` fields.
For consistency, only when in the onboarding wizard context, consider to make these two fields have the same width of the other fields.

afercia:
Fixes #96 

Uses 
`.yoast-wizard-text-input [type="text"]`
instead of
`.yoast-wizard-text-input-field`

to target any text input field inside its container, regardless of the input CSS class.

### DIFF
--- a/composites/OnboardingWizard/onboarding-wizard.scss
+++ b/composites/OnboardingWizard/onboarding-wizard.scss
@@ -221,7 +221,7 @@ a:focus {
       font-weight: bold;
     }
 
-    &-field {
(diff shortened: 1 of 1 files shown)


rensw90:
CR + Acceptance done 👍

PR #102, opened 2016-10-18
Change request type.

andizer:
Change the request type from PUT into POST, because not every webserver has PUT enabled.


atimmer:
CR Done 👍🏻
Acceptance done 👍🏻

Issue #114, opened 2016-11-16
When switching between between different text inputs with a conditional values are kept


atimmer:
When you use the following configuration:
```
{
	"fields": {
		"welcomeService": {
			"componentName": "HTML",
			"properties": { "html": "<p>Welcome to the Yoast Local SEO configuration wizard. In a few simple steps we'll help you configure your Local SEO settings to match your website's needs!<\/p>" }
		},
		"singleOrMultiple": {
			"componentName": "Choice",
			"properties": {
				"choices": {
					"single": { "label": "One single location" },
					"multiple": { "label": "Multiple locations" }
				},
				"label": "Please specify if you will be using one single location or multiple locations (Yoast SEO will create a new Custom Post Type from you where you can manage your locations)."
			},
			"data": "single"
		},
		"conditional": {
			"componentName": "Choice",
			"properties": {
				"choices": {
					"first": { "label": "First choice" },
					"second": { "label": "Second choice" }
				},
				"label": "Choose and we will do thing conditionally."
			},
			"data": "single"
		},
		"choiceFirst": {
			"componentName": "Input",
			"properties": { "label": "First input" },
			"requires": { "field": "conditional", "value": "first" },
			"data": "First prefilled"
		},
		"choiceSecond": {
			"componentName": "Input",
			"properties": { "label": "Second input" },
			"requires": { "field": "conditional", "value": "second" },
			"data": "Second prefilled"
		},
		"successMessage": {
			"componentName": "HTML",
			"properties": { "html": "<p>You\u2019ve done it! You've set up Yoast Local SEO<\/p>" }
		},
	},
	"steps": {
		"intro": {
			"title": "Welcome!",
			"fields": [ "welcomeService", "mailchimpSignup" ]
		},
		"singleOrMultipleLocations": {
			"title": "Single or multiple locations?",
			"fields": [ "conditional", "choiceFirst", "choiceSecond" ]
		}
	}
}
```

As far as I can see https://github.com/Yoast/yoast-components/commit/f6e40c0402e903979977f4832bb1d2ce2684745b introduced a key to the input fields, but this isn't enough. When changing `${key}` to `${name}` this bug is solved.

PR #123, opened 2016-12-05
Select added as an option to the Choice component + plus a little bugfix

jester1979:
You can now add `$this->set_property( 'type', 'select' );` to the `__constructor()` of your `WPSEO_Config_Field_Choice` class. The choices you added will now be represented as a regular selectbox ( instead of radios )

Bugfix
Give each field a unique fieldkey based upon its name, rather than a key.

Fixes: #114

### DIFF
--- a/composites/OnboardingWizard/components/Choice.js
+++ b/composites/OnboardingWizard/components/Choice.js
@@ -20,9 +20,10 @@ const Choice = ( props ) => {
 
 	let fieldSet = () => {
 		let choiceName;
+		let field;
 
 		if ( type === "select" ) {
-			return (
+			field =
 				<fieldset className={"yoast-wizard-input-select-" + fieldName}>
 					<select defaultValue={props.value} name={fieldName}
 					        className={props.optionClassName + " " + choiceName} onChange={props.onChange}>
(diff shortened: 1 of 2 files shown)


composites/OnboardingWizard/components/Choice.js
```diff
@@ -5,8 +5,8 @@ import htmlDecoder from "../helpers/htmlDecoder";
 import Explanation from "./Explanation";
 
 /**
- * Represents a choice interface, like a group of radio buttons or a select button. Initially it should render a
- * group of radio buttons. We might add other representations later on.
+ * Represents a choice interface, like a group of radio buttons or a select button. It could render a
+ * group of radio buttons (default) or a selectbox
```
terw-dan: Missing period at the end of the sentence.

composites/OnboardingWizard/components/Choice.js
```diff
@@ -16,29 +16,61 @@ const Choice = ( props ) => {
...
+							return (
+								<option value={choiceName} key={index}>
+									{htmlDecoder( choice.label )}
+								</option>
+							);
+						} )}
+					</select>
+				</fieldset>
+			;
+		} else {
```
terw-dan: Instead of an else, I think it would be better to do an early return.

terw-dan:
Closing this in favor of #124

PR #127, opened 2017-01-13
Removes width from section

terw-dan:
This removes the fixed width from the section to make sure we can use `yoast-section` responsive.


jcomack:
CR + Acceptance done 👍

PR #138, opened 2017-03-17
Stories/bl/add yellow

boblinthorst:
Added a yellow color needed for my yoast.
https://github.com/Yoast/my-yoast/issues/9.


afercia:
Was hoping to discuss a bit colors naming conventions, but then I went ahead. I'd propose to name colors with as generic as possible names. It doesn't make much sense to me to create multiple alias mapped to the same colors just to have a name that fits a specific use case. For example"
```
$background:                 $palette_grey_light;
$input_my_yoast:             $palette_grey_light;
```

Could be just one `$color_grey_light`. Any feedback welcome.

Updated style guide comment for the new yellow.
CR 👍

afercia:
Chatted a bit with @atimmer and decided it makes more sense to use more generic names for the color aliases. Changed the new yellow alias in just `$color_yellow`.
Other aliases names like `nav_my_yoast`, `input_my_yoast`, `zebra_stripe` etc. should be addressed in a separate issue. Worth noting we should also try to keep the .json file as small as possible :)

Issue #90, opened 2016-09-20
Make advanced select component


andizer:
We want a select component to be searchable.

[React Select](http://jedwatson.github.io/react-select) Is the most used library but it is not accessible.

We could consider forking that one or look for another react library, or build something ourselves.

afercia:
Had a quick look at React Select and at least it uses the main ARIA roles correctly :) Checked just the first example though. The reason why the options are not announced by VoiceOver it's the wrong usage of `aria-selected="true"` that should be set dynamically on the hovered/focused options. The options get announced on Windows, both with Firefox+NVDA and IE11+JAWS. 
There are several details to fix though and the advanced examples may need a lot of work.

![react-select](https://cloud.githubusercontent.com/assets/1682452/18679661/f238fed6-7f60-11e6-85cb-143c27d0e314.png)

atimmer:
We can actually use select2 here using a select2-react-wrapper. Let's use https://github.com/rkit/react-select2-wrapper. We are already using select2 in other places.

PR #140, opened 2017-03-21
Rename color aliases.

### LINKED ISSUE #139: Make the color names aliases less specific
Splitting this out from #138

It makes sense to use more generic names for the color aliases. Names like `nav_my_yoast`, `input_my_yoast`, `zebra_stripe` etc. are far too specific. Worth noting we should also try to keep the .json file as small as possible and avoid duplicating aliases mapped to the same color.

Once done, some of the colors used on My Yoast component should be updated.

afercia:
Rename aliases with less specific names. See also https://github.com/Yoast/my-yoast/issues/89

```
removed
$color_button_my_yoast
$color_link_my_yoast
$color_toggle_on_knob
because there's already $color_blue

removed
$color_toggle_on_gutter
there's already $color_blue_light

removed
$nav_my_yoast
there's already $color_pink_dark

removed
$zebra_stripe
there's already $color_background_light

removed
$background and $input_my_yoast
now it's $color_grey_light
```

Fixes #139


style-guide/_colors.scss
```diff
@@ -77,16 +77,9 @@ $color_grey_quote:           $palette_grey_text;
 $color_grey_text:            $palette_grey_text;
 
 // My Yoast colors.
-$color_chevron:              $palette_grey_medium_dark;
-$color_round_button:         $palette_green_medium_light;
```
boblinthorst: I don't see round button replaced in #90. but it is removed here.

afercia: On https://github.com/Yoast/my-yoast/pull/90 I've replaced the colors currently used on master. The ones in the pending PRs should be updated on each PR before merging.

boblinthorst: wow, based on the names of some of these colors, I was convinced they were used in certain spots that were unedited. But It seems you were right. Looking through the master branch, they indeed aren't used in those spots. Makes me wonder why some of those were made in the first place.

PR #147, opened 2017-06-07
Update the screen-reader-text CSS class.

### LINKED ISSUE #146: Update the screen-reader-text CSS class
As done in other places, we should consider to update the `screen-reader-text` CSS class with more modern CSS properties to target modern browsers. Specifically, `clip` is non-standard and going to be replaced by `clip-path`.

afercia:
Update the `screen-reader-text` CSS class as done for My Yoast.

Fixes #146


IreneStr:
CR done :+1:

IreneStr:
Acceptance done 👍

PR #153, opened 2017-07-12
Will remove console errors caused by using React.PropTypes and React.createClass

Xyfi:
Will remove console errors caused by using React.PropTypes instead of prop-types package.

Before React would give a deprecation error cause by the use of React.Proptypes instead of the new `prop-types` package. This involved upgrading a few dependencies like `material-ui` and `i18n-calpyso`.


Xyfi:
Will also fix `React.createClass` error. In the process of fixing it I also updated webpack to version 3+.

composites/OnboardingWizard/tests/wizardTest.js
```diff
@@ -1,9 +1,13 @@
+// console.log( '???' );
```
thulshof: What's this?

composites/OnboardingWizard/tests/wizardTest.js
```diff
@@ -1,9 +1,13 @@
...
 jest.unmock( "../OnboardingWizard" );
 jest.unmock( "lodash/cloneDeep" );
 jest.unmock( "../config/production-config" );
 jest.unmock( "material-ui/utils/withWidth" );
 jest.unmock( "../../../utils/i18n" );
-jest.mock( '../helpers/ajaxHelper', () => {
+jest.unmock( "prop-types" );
+// jest.unmock( "[Symbol.iterator]" );
```
thulshof: Why is this commented?

composites/SearchResultForm/SearchResultForm.js
```diff
@@ -1,6 +1,7 @@
 import React from "react";
-import { localize } from "i18n-calypso";
+import PropTypes from "prop-types";
 
+import { localize } from "../../utils/i18n";
```
thulshof: Why have you added this? I don't see any new code related to `localize`?

Xyfi: I did a search and replace for all occurences of `import { localize } from "i18n-calypso";`, it doesn't seem necessary indeed.

thulshof:
CR :+1:

Xyfi:
I would like to point out that you still get `React.PropTypes` errors in `wordpress-seo`, but these are caused by it's own components.

IreneStr:
Acceptance 👌

moorscode:
@afercia could you create an issue for that ^

PR #172, opened 2017-07-25
Stories/ab/157 implement progress bar

### LINKED ISSUE #157: Implement ProgressBar Component
Using styled components, implement a `ProgressBar` component with the following props:
- minimumValue `number` The minimum value
- maximumValue `number` The maximum value
- currentValue `number` The current value

Don't forget to add `aria-hidden="true"` since we want to hide the progress bar for screen readers as long as we haven't found a way to announce it properly (see: https://github.com/Yoast/YoastSEO.js/issues/466)

schermafbeelding_2017-07-14_om_12_05_37

Please put this component in the SnippetPreview folder.

Xyfi:
Implements the progress bar component in react.

The progress bar is a react implementation of the progress bar in [Yoast/YoastSEO.js](https://github.com/Yoast/YoastSEO.js).

To test replace the contents of `SnippetPreview.js` with:
```js
import React from "react";

import ProgressBar from "./ProgressBar";

	height: 700px;
	width: 100%;
	background-color: white;
`;

/**
 *
 */
	return (
		<SnippetPreviewDiv>
			<ProgressBar max={ 100 } value={ 50 } />
		</SnippetPreviewDiv>
	);
}
```

Fixes #157


composites/SnippetPreview/components/ProgressBar.js
```diff
@@ -0,0 +1,56 @@
...
+import colors from "../../../style-guide/colors.json";
+
+const ProgressBarContainer = styled.progress`
+	box-sizing: border-box;
+	width: 100%;
+	height: 8px;
+	display: block;
+	margin-top: 5px;
+	border: none;
+	-webkit-appearance: none;
```
afercia: Please don't use prefixed CSS properties, they should be added during the build process

composites/SnippetPreview/components/ProgressBar.js
```diff
@@ -0,0 +1,56 @@
...
+	appearance: none;
+	
+	::-webkit-progress-bar {
+	   	background-color: ${colors.$color_background_light};
+	   	color: ${( { color } ) => color };
```
afercia: Line 17 and 18 have spaces used for indentation, please use only tabs.
Line 17: please add spaces inside brackets

composites/SnippetPreview/components/ProgressBar.js
```diff
@@ -0,0 +1,56 @@
...
+		border: 1px solid ${colors.$color_input_border};
```
afercia: please add spaces inside brackets

composites/SnippetPreview/components/ProgressBar.js
```diff
@@ -0,0 +1,56 @@
...
+	}
+	
+	::-webkit-progress-value {
+		background-color: ${ props => props.progressColor };
+		transition: width 250ms;
+	}
+`;
```
afercia: Seems the styling of the progress bar will work on webkit browsers only? We should support all browsers. Please refer to YoastSEO.js `_snippet-editor.scss` where there are some rules and a `@mixin` for the progress bar.

composites/SnippetPreview/components/ProgressBar.js
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * The ProgressBar component
+ *
+ * @summary
+ * A component that displays a progress bar
```
afercia: not sure why the summary goes in a new line

composites/SnippetPreview/components/ProgressBar.js
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * @param {object} props Component props
```
afercia: Please add a period at the end

composites/SnippetPreview/components/ProgressBar.js
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * @returns {ReactElement} The ProgressBar component.
+ */
+function ProgressBar( props ) {
+	return (
+		<ProgressBarContainer max={props.max} value={props.value} progressColor={props.progressColor} aria-hidden="true"/>
```
afercia: Please add spaces inside brackets

afercia:
CR done 🚧 a few minor coding style issues. About styling, probably needs more work to support all browsers?

Xyfi:
I've added browser specific styling. Keep in mind that the `progress` HTML5 element is not properly support in firefox, that's why it looks different in firefox and isn't animated. For more info see [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/::-moz-progress-bar).

IreneStr:
@xyfi Please fix the eslint errors and resolve the merge conflicts

App.js
```diff
@@ -6,6 +6,7 @@ import Wizard from "./composites/OnboardingWizard/OnboardingWizard";
 import Config from "./composites/OnboardingWizard/config/production-config";
 import SearchResultsEditor from "./composites/SearchResultEditor/SearchResultEditor";
 import SnippetPreview from "./composites/SnippetPreview/components/SnippetPreview";
+import ProgressBarPreview from "./composites/SnippetPreview/components/ProgressBarPreview";
```
IreneStr: Please don't commit ProgressbarPreview, because that is essentially an acceptance test. If we would commit a preview for every component, we would soon have a gazillion tabs/button in the app. Please provide test instructions in the PR message on how to test the ProgressBar in `components/SnippetPreview.js` instead. If necessary, you can include a code snippet that the tester can copy and paste.

composites/SnippetPreview/components/ProgressBar.js
IreneStr: Please refer to my-yoast for our documentation standards.

composites/SnippetPreview/tests/ProgressBarTest.js
```diff
@@ -0,0 +1,31 @@
...
+test( "the progress bar matches the snapshot", () => {
+	const component = renderer.create(
+		<ProgressBar/>
+	);
+
+	let tree = component.toJSON();
+	expect( tree ).toMatchSnapshot();
+} );
+
+test( "the progress bar with progress matches the snapshot", () => {
```
IreneStr: This test is kind of redundant because of the next test

composites/SnippetPreview/components/ProgressBar.js
```diff
@@ -0,0 +1,63 @@
...
+	display: block;
+	margin-top: 5px;
+	border: none;
+	-moz-appearance: none;
+	-webkit-appearance: none;
+	appearance: none;
+
+	::-webkit-progress-bar {
+	   	background-color: ${colors.$color_background_light};
+	   	color: ${( { progressColor } ) => progressColor };
```
afercia: there are spaces on line 18 and 19, please use tabs.
Also, not sure about the syntax `${( { progressColor } ) => progressColor }` and why below a different syntax is used: `${ props => props.progressColor }`. I'd use always the second one, also in the other cases.

composites/SnippetPreview/components/ProgressBar.js
```diff
@@ -0,0 +1,63 @@
...
+const ProgressBarContainer = styled.progress`
+	box-sizing: border-box;
+	width: 100%;
+	height: 8px;
+	display: block;
+	margin-top: 5px;
+	border: none;
+	-moz-appearance: none;
+	-webkit-appearance: none;
+	appearance: none;
```
afercia: styled-components processes the provided styles already adding prefixes where necessary. If we specify prefixed properties explicitly, the result is duplicated properties:

screen shot 2017-08-02 at 10 51 37

Please remove the prefixed properties.

composites/SnippetPreview/components/ProgressBar.js
```diff
@@ -0,0 +1,63 @@
...
+		border: 1px solid ${colors.$color_input_border};
+	}
+
+	::-webkit-progress-value {
+		background-color: ${ props => props.progressColor };
+		transition: width 250ms;
+	}
+
+	::-moz-progress-bar {
+		background-color: ${ ( { progressColor } ) => progressColor };
```
afercia: See syntax.

composites/SnippetPreview/components/ProgressBar.js
```diff
@@ -0,0 +1,63 @@
...
+		border: 1px solid ${colors.$color_input_border};
+	}
+	
+	::-ms-fill {
+		background-color: ${ ( { progressColor } ) => progressColor };
```
afercia: See syntax.

composites/SnippetPreview/components/ProgressBar.js
```diff
@@ -0,0 +1,63 @@
...
+	}
+}`;
+
+/**
+ * The ProgressBar component.
```
afercia: Not sure why we're removing `@summary`

afercia:
CR 3 done 🚧 
See comments. Also, I know it's largely inconsistent across yoast-components but as best practice for coding standards there should always be spaces between brackets. For example:
```
${colors.$color_input_border}
{props.max}
```

should be
```
${ colors.$color_input_border }
{ props.max }
```

This should be applied to the whole PR.

afercia:
Forgot to mention: can't test on IE / Edge, because stubby seems to fail and I get nothing to see at `http://localhost:3333`. Any suggestions welcome!

composites/SnippetPreview/components/ProgressBar.js
IreneStr: For consistency with the my-yoast project ;)

Xyfi:
The windows compatibility issue has been address in PR #188.

afercia:
Seems to me there's no feedback about https://github.com/Yoast/yoast-components/pull/172#discussion_r130830192
there are still spaces on those two lines. Maybe it's a problem with the editor used, I'd recommend to make the invisible characters always visible.
Also, no feedback about the different syntax.

package.json
```diff
@@ -27,6 +27,7 @@
   "dependencies": {
     "a11y-speak": "git+https://github.com/Yoast/a11y-speak.git#master",
     "algoliasearch": "^3.22.3",
+    "grunt-scss-to-json": "^1.0.1",
```
afercia: This is also in the optionalDepdendencies, I know @atimmer moved it there because of issues on Travis, not sure it should be specified in both the dependencies and the optionalDepdendencies.

afercia:
A couple issues on Windows.

IE 11: can't test because of a JS error:
`SCRIPT5022: Objects are not valid as a React child (found: object with keys {$$typeof, type, key, ref, props, _owner, _store}). If you meant to render a collection of children, use an array instead or wrap the object using createFragment(object) from the React add-ons. Check the render method of `RedBoxError`

ie11

Seems to me `redbox-react` is something used by `react-hot-loader`, we should investigate this issue.


Edge:
 
edge

Works, but there's no background color. We should check check the best way to provide  a cross-browsers background color. There's also a thin line on the right of the progress bar, no idea how to fix it.

afercia:
Update on the IE 11 issue: seems this fixes it for me https://github.com/facebook/react/issues/8379#issuecomment-316346239
really not my area of expertise, though. /cc @atimmer 

That said, the background issue is visible on IE 11 too:

screenshot 32

atimmer:
@afercia Ah yes, we need to load babel-polyfill in the App.js in `yoast-components`.

afercia:
@atimmer doesn't work for me. Works only if I add `babel-polyfill` also to `webpack.config.js` as first thing in `entry: []`.

A little, unrelated, mystery is why I have a `main.js` file on my macOS install and no `main.js` on my Windows install.

boblinthorst:
CR: seems fine to me. though there still seem to be issues with this pull, looking at the issues Andrea seems to be having so i'm not entirely comfortable moving it to code review done.

Xyfi:
@afercia @atimmer Adding it to entry is the right way when using webpack according to the documentation. Since webpack is now supported in `wordpress-seo` (as soon as the PR has come through), we should document what the webpack requirements (loaders, entry scripts, etc)  are for `yoast-components` in the `README.md`. That way we can keep the `wordpress-seo` webpack config up to date.

afercia:
To recap_
- the progress bar "track" has no background color in Edge and IE 11, see code example below
- the thin right border on Edge and IE 11 should be removed, see code example below
- still not sure why there's a different syntax for the background colors
- the `babel-polyfill` thing needs @atimmer's blessing 

I've tried this styling in Chrome, Firefox, Edge, and IE 11 on Windows. Used some bright colors to make testing easier:
```
const ProgressBarContainer = styled.progress`
	box-sizing: border-box;
	width: 100%;
	height: 8px;
	display: block;
	margin-top: 5px;
	border: none;
	appearance: none;
	background-color: aqua;
	border: 1px solid red;

	::-webkit-progress-bar {
		background-color: aqua;
	}

	::-webkit-progress-value {
		background-color: ${ props => props.progressColor };
		transition: width 250ms;
	}

	::-moz-progress-bar {
		background-color: ${ ( { progressColor } ) => progressColor };
	}

	::-ms-fill {
		background-color: ${ ( { progressColor } ) => progressColor };
		border: 0;
	}
}`;
```

Seems to me the result is consistent across all browsers tested. Please check it again in all browsers, and replace the colors with the right ones.

screenshot 41

screenshot 42

screenshot 43

screenshot 44

boblinthorst:
CR: Ok 👍

boblinthorst:
Acceptance: seems to work. Hope we covered our bases :)

PR #173, opened 2017-07-25
Implement generic button component

### LINKED ISSUE #165: Implement a generic Button component
Using styled components, implement a generic `Button` component with the following props:

- onClick `function` The function that is called when the button is clicked
- text `string` The text within the button

schermafbeelding_2017-07-14_om_12_05_37

Please put this component in the Plugin folder (the folder containing SnippetPreview and ContentAnalysis).

Xyfi:
Fixes #165


composites/SnippetPreview/components/Button.js
```diff
@@ -0,0 +1,34 @@
+import styled from "styled-components";
+
+/**
+ * @summary:
+ * Returns a basic styled button
```
afercia: Please don't split summary in two lines, remove the colon, and add a period at the end

composites/SnippetPreview/components/Button.js
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * @param {object} props Component props
```
afercia: Please add a period at the end

composites/SnippetPreview/components/Button.js
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * @returns {ReactElement} Styled button
```
afercia: Please add a period at the end

composites/SnippetPreview/components/Button.js
```diff
@@ -0,0 +1,34 @@
...
+    border: 1px solid #dbdbdb;
+    border-radius: 4px;
+    color: #555;
+    background: #f7f7f7;
+    font-size: 0.8rem;
+    line-height: 15px;
+    cursor: pointer;
+    box-sizing: border-box;
+    font-family: inherit;
+    font-weight: inherit;
```
afercia: all these lines use spaces for indentation, please use only tabs. Will add considerations about styling in the comments

composites/SnippetPreview/components/Button.js
```diff
@@ -0,0 +1,34 @@
...
+    
+    :focus {
+		border-color: #5b9dd9;
+		outline: none;
+		color: #23282d;
+		background-color: #fafafa;
+		box-shadow: 0 0 3px rgba(0, 115, 170, 0.8);
+    }
```
afercia: Please use tabs instead of spaces for indentation

afercia:
CR done 🚧 
Some minor issues to address. Some more important considerations:
- I'd consider to add a prop for the button `type` attribute, default `button`. Button elements can be of type submit, reset, button. With `type="button"` the default action is prevented by default, which is very handy when the button has a `click` event attached and allows to don't use `preventDefault()` because there's no default action by default
- styling is a bit more complicated, buttons are tricky

I think there's still the need to reset some Firefox specific styling. For example in WordPress this is done with:
```
/* Remove the dotted border on :focus and the extra padding in Firefox */
.wp-core-ui button::-moz-focus-inner,
.wp-core-ui input[type="reset"]::-moz-focus-inner,
.wp-core-ui input[type="button"]::-moz-focus-inner,
.wp-core-ui input[type="submit"]::-moz-focus-inner {
	border-width: 0;
	border-style: none;
	padding: 0;
}
```

Font size and line-height:
```
font-size: 0.8rem;
line-height: 15px;
```

This way, the font-size is relative to the root but then the line-height is fixed. This could lead to unexpected results. Also, with 16px as base size, the computed value would be `12.8px` which is a bit pointless. I'd consider to inherit the font-size.
Ideally, the line-height should be unitless.
I understand this is inspired by the snippet editor button, but the current implementation is basically a CSS hack and we should aim for a more flexible solution.

Focus style:
the current implementation replicates the WordPress focus style. However, this component can be reused on other platforms too. I think we should make our components as much platform-agnostic as possible. 
The current situation in the plugin and components is mixed at best. I'd consider to leave out the focus style, as this should be implemented at a platform integration level, i.e. provide the necessary styles in the code that makes use of the Button component.
For this reason, I'd consider to add a default className, something like `yoast-button`.

Alternately:
- make a base button with very basic styling
- extend the base button to make a button for WordPress, with WP specific styles

Any thoughts welcome!

composites/SnippetPreview/components/Button.js
IreneStr: Please refer to the my-yoast components for our documentation conventions ;)

Xyfi:
I have gone through the CR and have fixed the following:
* `Button` has been split up into a basic `Button` and a `SnippetPreviewButton` that inherits from `Button`
* Added specific fix for firefox (`-moz-focus-inner`)
* Replaced hard-coded colors by `colors.json` colors
* `Button` receive the following `defaultProps`
  * `className: ["button", "yoast-button"]`
  * `type: "button"`
* Removed `:focus` selector

I had some discussion with @afercia about how to handle styles in this repo, since some styles will be affected by the global styles in wordpress. We agreed we should discuss this in further detail with @atimmer.

composites/SnippetPreview/components/Button.js
```diff
@@ -0,0 +1,53 @@
...
+	font-family: inherit;
+	font-weight: inherit;
+
+	::-moz-focus-inner {
+		border-width: 0;
+		padding: 0;
+	}
+
+	:focus {
+		border-color: #5b9dd9;
```
atimmer: This should be from `_colors`.

composites/SnippetPreview/components/Button.js
```diff
@@ -0,0 +1,53 @@
...
+		outline: none;
+		color: #23282d;
+		background-color: #fafafa;
```
atimmer: See colors comment.

composites/SnippetPreview/components/Button.js
```diff
@@ -0,0 +1,53 @@
...
+	font-size: inherit;
+	font-family: inherit;
+	font-weight: inherit;
+
+	::-moz-focus-inner {
+		border-width: 0;
+		padding: 0;
+	}
+
+	:focus {
```
atimmer: We should create an `addFocusStyles` function which is used for this button and everywhere where we need this focus style.

atimmer: This can be used in `my-yoast` too 🎉

afercia:
After some research, it seems something changed recently in Firefox and they removed the extra padding that used to be applied on the buttons. I _think_ this is the relevant Bugzilla ticket: https://bugzilla.mozilla.org/show_bug.cgi?id=140562

Firefox used to make buttons a bit bigger, see FF 48 on the left, compared with 54 on the right (both on Windows). The second buttons in each row have the CSS fix that targets `::-moz-focus-inner,` :

screenshot 40

as you can see on the right, in FF 54 the buttons size is always the same.

However, on Windows, when buttons are focused there's still a default thin dotted line:

screenshot 37

to remove this dotted line, the only thing we need is `border-width: 0;` so we can remove `padding: 0;`.

afercia:
Was thinking in 99% of the cases we'd want a button with focus style. Right now, we should do every time during implementation something like this:
`const FocusedSnippetPreviewButton = focus( SnippetPreviewButton );`
or
`const FocusedButton = focus( Button );`

Maybe worth considering same renaming and export the buttons _with_ focus?

composites/SnippetPreview/components/Button.js
```diff
@@ -0,0 +1,62 @@
...
+`;
+
+/**
+ * Returns a component with applied focus styles
+ *
+ * @param {ReactElement} component The original component
+ *
+ * @returns {ReactElement} Component with applied focus styles
+ */
+export function focus( component ) {
```
afercia: I'd consider to rename this function to better describe what it does. On My Yoast, we're already using this kind of stuff, and the names are a bit more descriptive. For example:
```

disable()
makeButtonFullWidth()
makeResponsiveIconButton()
separatify()
ellipsify()
makeFullWidth()
```

composites/SnippetPreview/components/Button.js
```diff
@@ -0,0 +1,62 @@
...
+	return styled( component )`
+		&::-moz-focus-inner {
+			border-width: 0;
+			padding: 0;
```
afercia: See comment, with latest Firefox `padding: 0;` is no longer needed.

afercia:
CR 2 done, a few minor things to address.
Most of all, I'd consider some renaming. See comment above. Maybe start with a `BaseButton`?
I'd rename `focus()` to `addFocusStyle()`. And export the buttons _with_ focus.
Please also resolve the conflicts.

afercia:
Screenshot of the rendered button with focus:

screen shot 2017-08-07 at 13 27 45

afercia:
Agreed with @atimmer to implement and use a new helper function `rgba()` to generate and use rgba colors in the styled-components. Usage notes in the `style-guide/README.md` file.

Pending feedback on naming convention and placement of the new stuff.

Worth also noting colors prefixed with `$palette_` should not be used directly, please make an alias and use that instead.

Note: right now, the size and colors used for the button don't exactly match the ones of the WordPress buttons. This should be addressed after some design feedback, probably in a separate PR.

To test:
- switch to this branch
- replace the `composites/SnippetPreview/components/SnippetPreview.js` content with the content below, `yarn start`, go to http://localhost:3333/ and go in the Snippet preview page

```
import React from "react";
import { BaseButton, SnippetPreviewButton } from "./Button";

	height: 700px;
	width: 100%;
	text-align: center;
	background-color: white;
`;

/**
 *
 */
	return <SnippetPreviewDiv>
		<BaseButton>Base button</BaseButton><br/><br/>
		<SnippetPreviewButton>SnippetPreview button</SnippetPreviewButton>
	</SnippetPreviewDiv>;
}
```

IreneStr:
Acceptance done 👍

PR #185, opened 2017-08-02
Add inline comment for the new palette_red_light color

### LINKED ISSUE #184: Add comment for new color palette_red_light
See #178 

All colors in `_colors.scss` should have an inline comment to give _some_ indication about the other color they can be paired with to have a minimum color contrast ratio of 4.5:1

afercia:
Fixes #184

All colors in _colors.scss should have an inline comment to give some indication about the other color they can be paired with to have a minimum color contrast ratio of 4.5:1


Xyfi:
CR :+1: Acceptance :+1:

PR #218, opened 2017-08-09
Added Svg Icon component using svg-react-loader

### LINKED ISSUE #189: Implement Icon component
Using styled components, implement an Icon component with the following props:

- color `string` The icon's color.
- height `string` The icon's height
- icon `string` The url to the icon's svg

Please put this component in the Plugin folder (the folder containing SnippetPreview and ContentAnalysis).

Xyfi:
Also added jest configuration to map svg files to `/jest/__mocks__/svgReactLoaderMock.js`, because jest doesn't use webpack loaders.

## Technical decisions
* Implemented `svg-react-loader`, which creates react components from `.svg` files.
* Added svg mocking instructions to `package.json` that tell jest to mock imports of `.svg` files to `/jest/__mocks__/svgReactLoaderMock.js`.

## Testing
1. Paste the following in `SnippetPreview.js`:
```js
import React from "react";

import { Icon } from "./Icon";
import { edit } from "../../../style-guide/svg";

	height: 700px;
	width: 100%;
	background-color: white;
`;

/**
 *
 */
	return <SnippetPreviewDiv><Icon icon={ edit } iconColor="#000"/></SnippetPreviewDiv>;
}

```
2. Test using the `iconSize` and `iconColor` props on the `Icon` component.
3. Additionally you can add and `.svg` file (make sure it's a square icon) to `/style-guide/svg` and export it in `/style-guide/svg/index.js`. Then the name you use to export the svg in `index.js` can be used as the `icon` prop on the `Icon` component.

Fixes #189

### DIFF
--- a/composites/SnippetPreview/components/Icon.js
+++ b/composites/SnippetPreview/components/Icon.js
@@ -12,19 +12,20 @@ import styled from "styled-components";
 export const Icon = ( props ) => {
 	const IconComponent = styled( props.icon )`
 		float: left;
-		width: ${ props => props.size }px;
-		height: ${ props => props.size }px;
-		fill: ${ props => props.color };
+		width: ${ props => props.iconSize }px;
+		height: ${ props => props.iconSize }px;
+		fill: ${ props => props.iconColor };
 	`;
-	return <IconComponent aria-hidden="true" { ...props } />;
+
+	return <IconComponent aria-hidden="true" iconSize={ props.iconSize } iconColor={ props.iconColor } />;
 };
 
 Icon.propTypes = {
 	icon: PropTypes.func.isRequired,
-	color: PropTypes.string.isRequired,
-	size: PropTypes.number,
+	iconColor: PropTypes.string.isRequired,
+	iconSize: PropTypes.number,
 };
 
 Icon.defaultProps = {
-	size: 16,
+	iconSize: 16,
 };
--- a/composites/SnippetPreview/components/Icon.js
+++ b/composites/SnippetPreview/components/Icon.js
@@ -20,7 +20,7 @@ export const Icon = ( props ) => {
 };
 
 Icon.propTypes = {
-	icon: PropTypes.string.isRequired,
+	icon: PropTypes.func.isRequired,
 	color: PropTypes.string.isRequired,
 	size: PropTypes.number,
 };
--- a/composites/SnippetPreview/components/Icon.js
+++ b/composites/SnippetPreview/components/Icon.js
@@ -16,7 +16,7 @@ export const Icon = ( props ) => {
 		height: ${ props => props.size }px;
 		fill: ${ props => props.color };
 	`;
-	return <IconComponent { ...props } />;
+	return <IconComponent aria-hidden="true" { ...props } />;
 };
 
 Icon.propTypes = {
--- a/composites/SnippetPreview/components/Icon.js
+++ b/composites/SnippetPreview/components/Icon.js
@@ -2,17 +2,15 @@ import React from "react";
 import PropTypes from "prop-types";
 import styled from "styled-components";
 
-import * as Icons from "../../../style-guide/svg";
-
 /**
- * Returns an icon.
+ * Returns the Icon component.
  *
  * @param {object} props Component props.
  *
- * @returns {ReactElement} Icon.
+ * @returns {ReactElement} Icon component.
  */
 export const Icon = ( props ) => {
-	const IconComponent = styled( Icons[ props.icon ] )`
+	const IconComponent = styled( props.icon )`
 		float: left;
 		width: ${ props => props.size }px;
 		height: ${ props => props.size }px;
@@ -22,13 +20,11 @@ export const Icon = ( props ) => {
 };
 
 Icon.propTypes = {
-	icon: PropTypes.string,
-	color: PropTypes.string,
+	icon: PropTypes.string.isRequired,
+	color: PropTypes.string.isRequired,
 	size: PropTypes.number,
 };
 
 Icon.defaultProps = {
 	size: 16,
-	color: "#555",
-	icon: "edit",
 };
--- a/composites/SnippetPreview/components/Icon.js
+++ b/composites/SnippetPreview/components/Icon.js
@@ -1,0 +1,34 @@
+import React from "react";
+import PropTypes from "prop-types";
+import styled from "styled-components";
+
+import * as Icons from "../../../style-guide/svg";
+
+/**
+ * Returns an icon.
+ *
+ * @param {object} props Component props.
+ *
+ * @returns {ReactElement} Icon.
+ */
+export const Icon = ( props ) => {
+	const IconComponent = styled( Icons[ props.icon ] )`
+		float: left;
+		width: ${ props => props.size }px;
+		height: ${ props => props.size }px;
+		fill: ${ props => props.color };
+	`;
+	return <IconComponent { ...props } />;
+};
+
+Icon.propTypes = {
+	icon: PropTypes.string,
+	color: PropTypes.string,
+	size: PropTypes.number,
+};
+
+Icon.defaultProps = {
+	size: 16,
+	color: "#555",
+	icon: "edit",
+};
--- a/composites/SnippetPreview/components/Icon.js
+++ b/composites/SnippetPreview/components/Icon.js
@@ -11,21 +11,20 @@ import styled from "styled-components";
  */
 export const Icon = ( props ) => {
 	const IconComponent = styled( props.icon )`
-		float: left;
-		width: ${ props => props.iconSize }px;
-		height: ${ props => props.iconSize }px;
-		fill: ${ props => props.iconColor };
+		width: ${ props.iconSize };
+		height: ${ props.iconSize };
+		fill: ${ props.iconColor };
 	`;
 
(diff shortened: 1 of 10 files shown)


Xyfi:
Asked @atimmer to make sure mocking files that are handled by `webpack` during testing is desirable.

composites/SnippetPreview/components/Icon.js
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Returns an icon.
+ *
+ * @param {object} props Component props.
+ *
+ * @returns {ReactElement} Icon.
+ */
+export const Icon = ( props ) => {
+	const IconComponent = styled( Icons[ props.icon ] )`
```
atimmer: I think it might be better to pass the actual icon in with the props. Now `props.icon` is a string which means that it is hard to see where a certain icon has been used.

If you do it like this, it is clearer:
```
import edit from "[path-to-icons]/edit.svg";

// (...)

renderFunc() {
     return <Icon icon={ edit } />;
}
```

composites/SnippetPreview/components/Icon.js
```diff
@@ -0,0 +1,34 @@
+import React from "react";
+import PropTypes from "prop-types";
+import styled from "styled-components";
+
+import * as Icons from "../../../style-guide/svg";
+
+/**
+ * Returns an icon.
```
IreneStr: Returns the Icon component.

composites/SnippetPreview/components/Icon.js
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * @param {object} props Component props.
+ *
+ * @returns {ReactElement} Icon.
```
IreneStr: The Icon component.

composites/SnippetPreview/components/Icon.js
```diff
@@ -0,0 +1,34 @@
...
+Icon.propTypes = {
+	icon: PropTypes.string,
+	color: PropTypes.string,
+	size: PropTypes.number,
+};
+
+Icon.defaultProps = {
+	size: 16,
+	color: "#555",
+	icon: "edit",
```
IreneStr: It seems random that this specific SVG in this specific color is the default. I think it would be better to make color and icon required, and remove the defaults.

composites/SnippetPreview/tests/__snapshots__/IconTest.js.snap
```diff
@@ -0,0 +1,11 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`the Icon matches the snapshot 1`] = `
+<svg
+  aria-hidden="true"
+  className="sc-bdVaJa jEZkCV"
+  color="black"
+  icon={[Function]}
```
IreneStr: Function?

composites/SnippetPreview/tests/__snapshots__/IconTest.js.snap
```diff
@@ -3,9 +3,8 @@
 exports[`the Icon matches the snapshot 1`] = `
 <svg
   aria-hidden="true"
-  className="sc-bdVaJa jEZkCV"
-  color="black"
-  icon={[Function]}
-  size={16}
+  className="sc-bdVaJa inStFx"
+  iconColor={undefined}
```
IreneStr: Undefined?

IreneStr:
CR done :+1:

afercia:
Please always make sure to update the testing instructions when they're outdated because of following changes.

composites/SnippetPreview/components/Icon.js
```diff
@@ -0,0 +1,31 @@
...
+ */
+export const Icon = ( props ) => {
+	const IconComponent = styled( props.icon )`
+		float: left;
+		width: ${ props => props.iconSize }px;
+		height: ${ props => props.iconSize }px;
+		fill: ${ props => props.iconColor };
+	`;
+
+	return <IconComponent aria-hidden="true" iconSize={ props.iconSize } iconColor={ props.iconColor } />;
```
afercia: I'm getting `Warning: Unknown props `iconSize`, `iconColor` on <svg> tag. `
because those props are applied directly to the svg element and interpreted as non-standard attributes. `IconComponent` is actually a SVG element, we should probably use attributes width, height, fill, and pass them the values.

composites/SnippetPreview/components/Icon.js
```diff
@@ -0,0 +1,31 @@
...
+ * @param {object} props Component props.
+ *
+ * @returns {ReactElement} Icon component.
+ */
+export const Icon = ( props ) => {
+	const IconComponent = styled( props.icon )`
+		float: left;
+		width: ${ props => props.iconSize }px;
+		height: ${ props => props.iconSize }px;
+		fill: ${ props => props.iconColor };
```
afercia: In these 3 lines, we're using the props of the styled component. Why not use the props from the parent?

composites/SnippetPreview/components/Icon.js
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * Returns the Icon component.
+ *
+ * @param {object} props Component props.
+ *
+ * @returns {ReactElement} Icon component.
+ */
+export const Icon = ( props ) => {
+	const IconComponent = styled( props.icon )`
+		float: left;
```
afercia: Not sure it should be floated left by default. Are we assuming just one specific usage?

composites/SnippetPreview/components/Icon.js
```diff
@@ -0,0 +1,31 @@
...
+		fill: ${ props => props.iconColor };
+	`;
+
+	return <IconComponent aria-hidden="true" iconSize={ props.iconSize } iconColor={ props.iconColor } />;
+};
+
+Icon.propTypes = {
+	icon: PropTypes.func.isRequired,
+	iconColor: PropTypes.string.isRequired,
+	iconSize: PropTypes.number,
```
afercia: Can we make iconSize a string? Right now, to actually use this component you have to use this syntax:
` icon={ edit } iconColor="#f00" iconSize={ 40 }`
and remember to pass color as string but size as number. Not the most intuitive and easy to remember syntax.

composites/SnippetPreview/components/Icon.js
```diff
@@ -0,0 +1,31 @@
...
+};
+
+Icon.defaultProps = {
+	iconSize: 16,
```
afercia: Same: can this be a string?

style-guide/svg/edit.svg
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" standalone="no"?>
+<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN"
+        "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
```
afercia: Do we really need the xml prolog and the DTD? Asking because I relly don't know, but seems weird to me.

style-guide/svg/edit.svg
```diff
@@ -0,0 +1,6 @@
...
+<svg fill="#000" width="1792" height="1792" viewBox="0 0 1792 1792" xmlns="http://www.w3.org/2000/svg">
```
afercia: Any reason why the default size and box are 1792?

composites/SnippetPreview/tests/IconTest.js
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+import renderer from "react-test-renderer";
+
+import Edit from "../../../style-guide/svg/edit.svg";
```
afercia: Minor: "Edit" is not a component, and should be lowercase.

afercia:
Acceptance 🚧 
Some considerations to address, see inline comments.

On the implementation side: to my understanding, as it works now this would require to import each icon one by one. Say in a component we need to use 10 icons, then I'd need to:
```
import icon1 from "../../../style-guide/svg/icon1"
import icon2 from "../../../style-guide/svg/icon2"
import icon3 from "../../../style-guide/svg/icon3"
...
```

pointing this out just to be sure we're ok with this. /cc @atimmer

atimmer:
That's ok, if this gets too cumbersome, we can always create an `icons` file. In that case we can import like this:

```js
import { icon1, icon2, icon3 } from "../../../style-guide/svg/icons"
```

Xyfi:
There already was an `index.js` in the svg folder. So you can do: `import { icon1, icon2, icon3 } from "../../../style-guide/svg"`

style-guide/svg/edit.svg
Xyfi: I've lookedat the following document for how to define a svg file: https://www.w3.org/TR/SVG/struct.html

style-guide/svg/edit.svg
Xyfi: Yes, that is the canvas size the current icons are drawn on as defined in `css/_icons.css`

composites/SnippetPreview/tests/IconTest.js
Xyfi: Technically it is a component since we use `svg-react-loader`, but for clarity i agree lowercase is better.

boblinthorst:
CR and acceptance: Code seems fine. And the acceptance test shows it working the way i'd expect it to work. I hope I'm not missing anything when merging this one.

PR #223, opened 2017-08-14
Allow non-Icon specific props to be passed down in Icon component

Xyfi:
When applying a `styled-components` decorator to apply some additional stying to the Icon component I found out the changes were not showing up.

## Technical decisions
This PR fixes that by passing all props that are not specific for the Icon down to the outputted component. By using `lodash`'s `omit` all props that are not specific to the Icon component are removed and passed to the returned component.

This also allows the Icon's API to be cleaner by changing `iconSize` and `iconColor` to `size` and `color` respectively, because they won't be passed down to the html element as attributes anyway.

## Test
To test replace the content of `SnippedPreview.js` with the following content:

```js
import React from "react";

import { Icon } from "./Icon";
import { edit } from "../../../style-guide/svg";

	height: 700px;
	width: 100%;
	background-color: white;
`;

/**
 *
 */
	const DecoratedIcon = styled( Icon )`
		border: 10px solid black;
	`;
	return <SnippetPreviewDiv><DecoratedIcon icon={ edit } color="#000"/></SnippetPreviewDiv>;
};
```

### DIFF
--- a/composites/SnippetPreview/components/Icon.js
+++ b/composites/SnippetPreview/components/Icon.js
@@ -1,6 +1,7 @@
 import React from "react";
 import PropTypes from "prop-types";
 import styled from "styled-components";
+import _omit from "lodash/omit";
 
 /**
  * Returns the Icon component.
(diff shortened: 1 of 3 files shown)


boblinthorst:
Cr: ok 👍

boblinthorst:
Acceptance 👍

PR #225, opened 2017-08-15
Make button consistent with metabox wp buttons

### LINKED ISSUE #201: Make Snippet Preview buttons consistent with other metabox/wordpress buttons
Currently the button component styling in `wordpress-seo`, which this library's button is styled after, has some issues:

* It behaves differently on chrome and firefox/safari, which has to do with the `focus` behaviours in different browsers, not sure whether this can be fixed
* Could use some `hover` and `active` specific styles to emphasise it's a button, like most default wordpress buttons

Since we are rebuilding the Snippet preview from scratch, it would be a good idea to look into redesigning this component.

Xyfi:
This PR adds `:active` and `:hover` styles to the `BaseButton` to make it feel more like a button.

The new `BaseButton` is styled after the `Upload Image` button located in the Yoast SEO plugin under `Social > Facebook`.

## Test
To test replace the contents of `SnippetPreview.js` with the following content:
```js
import React from "react";
import { SnippetPreviewButton } from "./Button";

	height: 700px;
	width: 100%;
	background-color: white;
`;

/**
 *
 */
	return <SnippetPreviewDiv><SnippetPreviewButton>Text</SnippetPreviewButton></SnippetPreviewDiv>;
}
```

Fixes #201


IreneStr:
CR and acceptance done 👍  I can't test with IE and Edge at the moment. Maybe @afercia can take a look when he is back from his holiday.

PR #233, opened 2017-08-16
Implement analysisCollapsible component

### LINKED ISSUE #192: Implement AnalysisHeader component
Using styled components, implement an AnalysisHeader component with the following props:

- isOpen `bool` Whether the header is collapsed
- headerTitle `string` The segment title.
- children ` PropTypes.oneOfType( [
		PropTypes.arrayOf( PropTypes.node ),
		PropTypes.node,
	] )` The list items

You can use the collapsible header in the My Yoast project as an source of inspiration: `my-yoast/client/src/components/CollapsibleHeader.js`.

NB: this issue may change based on the discussion in https://github.com/Yoast/roadmap/issues/4

boblinthorst:
Implements issue #192.

Tecnical decisions:
Added an ID prop to ensure content always properly points to it's own title which can collapse it. This was no issue in my-yoast, where accessibility still needs improvements. 

Test instructions:
1. Checkout this branch. 
2. Replace the contents of `SnippetPreview.js` with the following code:

```
import React from "react";
import AnalysisCollapsible from "../../ContentAnalysis/components/AnalysisCollapsible";
import { IconButton } from "../../Shared/components/Button";

const SnippetPreviewDiv = styled.div`
	height: 700px;
	width: 80%;
	margin: 0 auto;
	background-color: white;

	button{
		margin-right: 10px;
	}

	.with-max-width {
		max-width: 152px;
	}
`;


export default class SnippetPreview extends React.Component {
	/**
	 * The constructor.
	 *
	 * @constructor
	 *
	 * @param {Object} props The props to use.
	 */
	constructor( props ) {
		super( props );
	}

	render() {
		return (
			<SnippetPreviewDiv>
				<h2>Buttons</h2>

				<h2> Analysis results </h2>
				<AnalysisCollapsible title="Problems">
					<li>Is gonna be alright.</li>
				</AnalysisCollapsible>
				<AnalysisCollapsible title="Problems">
				</AnalysisCollapsible>
				<AnalysisCollapsible title="Problems">
					<li>Is gonna be alright.</li>
					<li>About a thing.</li>
					<li>Cause every little thing,</li>
					<li>Is gonna be alright.</li>
				</AnalysisCollapsible>
			</SnippetPreviewDiv>
		);
	}
}
```

fixes #192


composites/Plugin/ContentAnalysis/components/AnalysisHeader.js
```diff
@@ -0,0 +1,151 @@
...
+			isOpen: ! this.state.isOpen,
+		} );
+	}
+
+	/**
+	 * Gets the correct arrow based on whether the list is collapsed or not.
+	 *
+	 * @returns {ReactElement} The upArrow when the header is collapsed, otherwise the downArrow.
+	 */
+	getArrow() {
```
Xyfi: Functions that help the main render function to render the component should have the render prefix, so in this case `renderArrow`.

composites/Plugin/ContentAnalysis/components/AnalysisHeader.js
```diff
@@ -0,0 +1,151 @@
...
+		let upArrow = <Icon icon={ angleUp } color= { colors.$color_grey_dark } size="20px" />;
```
Xyfi: 2 Components are being rendered and one is returned, this should be optimized.

composites/Plugin/ContentAnalysis/components/AnalysisHeader.js
```diff
@@ -0,0 +1,151 @@
...
+			children = this.props.children;
+		}
+
+		let childrenAmount = this.props.children.props.children.length;
+
+		return (
+			<AnalysisHeaderContainer>
+				<div role="presentation">
+					<h3 role="heading" aria-level="3">
+						<AnalysisHeader aria-expanded={ this.isOpen() } aria-controls="sect1" onClick={ () => {
```
Xyfi: `onClick` can be shortened to `onClick={ this.toggleOpen }`.

composites/Plugin/ContentAnalysis/components/AnalysisHeader.js
Xyfi: Space between `color=` and `{` should be removed.

composites/Plugin/ContentAnalysis/components/AnalysisHeader.js
```diff
@@ -0,0 +1,151 @@
...
+			isOpen: this.props.isOpen,
+		};
+
+		this.toggleOpen = this.toggleOpen.bind( this );
+	}
+
+	/**
+	 * Returns whether the list is collapsed.
+	 *
+	 * @returns {Boolean} True when the list is collapsed.
```
Xyfi: **False** when the list is collapsed.

From a programmer's standpoint using `collapsed` instead of `isOpen` would be clearer, maybe consider renaming the `isOpen` state variable.

Xyfi:
CR 🚧

composites/Plugin/ContentAnalysis/components/AnalysisCollapsible.js
```diff
@@ -0,0 +1,114 @@
+import PropTypes from "prop-types";
+import React from "react";
+import styled from "styled-components";
+import { angleUp, angleDown } from "../../../../style-guide/svg";
+import colors from "../../../../style-guide/colors.json";
+import defaults from "../../../../config/defaults.json";
+import { Icon } from "../../Shared/components/Icon";
+
+/**
+ * Container for the Collapsable header and it's content.
```
afercia: typos: Collapsable > Collapsible and it's > its

composites/Plugin/ContentAnalysis/components/AnalysisCollapsible.js
```diff
@@ -0,0 +1,114 @@
...
+	line-height: 1.33333333;
+	// Looks like Safari 10 doesn't like align-items: center for SVGs and needs some help.
+	align-self: center;
+`;
+
+/**
+ * A collapsible header used to show sets of analysis results. Expects list items as children.
+ *
+ * @param {object} props The properties for the component.
+ * @returns {XML} A collasible analysisresult set.
```
afercia: I think this should be {ReactElement} ?

composites/Plugin/ContentAnalysis/components/AnalysisCollapsible.js
```diff
@@ -0,0 +1,114 @@
...
+ * @constructor
```
afercia: Not sure this is correct.

composites/Plugin/ContentAnalysis/components/AnalysisCollapsible.js
```diff
@@ -0,0 +1,114 @@
...
+ */
+const AnalysisCollapsible = ( props ) => {
+	const headerContentId = props.headerId + "Content";
```
afercia: I think we don't need an ID, for the reasons I'll try to explain in a more expanded comment

composites/Plugin/ContentAnalysis/components/AnalysisCollapsible.js
```diff
@@ -0,0 +1,114 @@
...
+	return (
+		<AnalysisHeaderContainer>
+			<div role="presentation">
```
afercia: please remove role="presentation"

composites/Plugin/ContentAnalysis/components/AnalysisCollapsible.js
```diff
@@ -0,0 +1,114 @@
...
+				<h3 role="heading" aria-level="3">
```
afercia: please remove  role="heading" aria-level="3"
OR (to discuss) if we want the ability to specify the heading level via props, this should be a div with  role="heading" aria-level="3"

composites/Plugin/ContentAnalysis/components/AnalysisCollapsible.js
```diff
@@ -0,0 +1,114 @@
...
+					<AnalysisHeaderButton aria-expanded={ props.isOpen } aria-controls={ headerContentId }
```
afercia: please remove aria-controls and id (will try to expand more in a comment)

composites/Plugin/ContentAnalysis/components/AnalysisCollapsible.js
```diff
@@ -0,0 +1,114 @@
...
+										  id={ props.headerId } onClick={ props.onOpen }>
+						<AnalysisHeaderIcon icon={ props.isOpen ? angleUp : angleDown } color={ colors.$color_grey_dark } size="20px" />
+						<AnalysisTitle> { props.title + " (" + props.children.length + ")" } </AnalysisTitle>
+					</AnalysisHeaderButton>
+				</h3>
+			</div>
+			<ul id={ headerContentId } role="region" aria-labelledby={ props.headerId }>
```
afercia: please remove id, role, and aria-labelledby:
- id is not needed 
- not sure why we'r eusing role region on a list, this [overrides the native semantics](https://www.w3.org/TR/using-aria/#second) 
- a ul is not a [labelable element](https://www.w3.org/TR/html5/forms.html#category-label)

composites/Plugin/ContentAnalysis/tests/AnalysisHeaderTest.js
```diff
@@ -0,0 +1,37 @@
...
+				<li> Second item </li>
+				<li> Third item </li>
+				<li> Fourth item </li>
+			</ul>
+		</AnalysisCollapsible>
+	);
+
+	let tree = component.toJSON();
+	expect( tree ).toMatchSnapshot();
+} );
```
afercia: please update the test accordingly to the changes above

style-guide/svg/angle-down.svg
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" standalone="no"?>
+<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN"
+        "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
+<svg fill="#000" width="1792" height="1792" viewBox="0 0 1792 1792" xmlns="http://www.w3.org/2000/svg">
```
afercia: do we need the XML prolog and doctype?

style-guide/svg/angle-up.svg
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" standalone="no"?>
+<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN"
+        "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
```
afercia: same, do we need the XML prolog and doctype?

composites/Plugin/ContentAnalysis/components/AnalysisCollapsible.js
```diff
@@ -0,0 +1,114 @@
...
+			{ props.isOpen ? props.children : "" }
```
afercia: minor issue: indentation

afercia:
CR done 🚧 
Some issues about general usage of aria roles and attributes:

`role="presentation"` is meant to reset the native semantics of an element and its required children; it is _not_ inherited by elements that are not required children. For example. aet on a table element, it will reset also the tr, th, td semantics. However, set on a div it doesn't do anything because a div has no native semantics and no required children

- `<h3 role="heading" aria-level="3">` just repeats the native semantics of h3. No need to set them, as `role="presentation"` set on the wrapper div doesn't affect the heading semantics and we're removed it anyways. Instead, if we want the heading level to be dynamycally set via props, we could do something like: `<div role="heading" aria-level={ props.headingLevel }>`

- `aria-controls` could be of some (limited) usefulness if it only worked; see http://www.heydonworks.com/article/aria-controls-is-poop 🙂  I'd recommend to just don't use it. As a consequence e don't need `headerContentId`

-  UL element can't have a label or an aria-label, or ria-labelledby attribute (it doesn't do anything) so we don't need `props.headerId`

- please never reset an element native semantics: setting `role="region"` on the UL prevents assistive technologies from announcing the element as a list and it goes against the ARIA rules, see https://www.w3.org/TR/using-aria/#second

Side note: some of the styles used here, especially font size and line-height values, will actually depend on the font metrics used in the implementation; I guess values like `font-size: 1.2em;` and `line-height: 1.33333333;` were crafted on the My Yoast specific case and don't make much sense in a more abstracted context. Maybe we should consider to use more generic values or just inherit them, but this can also be improved later.

composites/Plugin/ContentAnalysis/components/AnalysisCollapsible.js
```diff
@@ -0,0 +1,114 @@
...
+			</ul>
+		</AnalysisHeaderContainer>
+	);
+};
+
+AnalysisCollapsible.propTypes = {
+	headerId: PropTypes.string.isRequired,
```
afercia: Based on the changes to implement, this will be no longer needed

composites/Plugin/ContentAnalysis/components/AnalysisCollapsible.js
```diff
@@ -0,0 +1,110 @@
...
+			<div>
+				<h3>
+					<AnalysisHeaderButton aria-expanded={ props.isOpen } onClick={ props.onOpen }>
+						<AnalysisHeaderIcon icon={ props.isOpen ? angleUp : angleDown } color={ colors.$color_grey_dark } size="20px" />
+						<AnalysisTitle> { props.title + " (" + props.children.length + ")" } </AnalysisTitle>
+					</AnalysisHeaderButton>
+				</h3>
+			</div>
+			<ul >
+				{ props.isOpen ? props.children : "" }
```
afercia: Quickly discussed with @Xyfi and maybe in cases like this one we would use `null` instead of an empty string.

composites/Plugin/ContentAnalysis/components/AnalysisCollapsible.js
```diff
@@ -0,0 +1,110 @@
...
+		<AnalysisHeaderContainer>
+			<div>
+				<h3>
+					<AnalysisHeaderButton aria-expanded={ props.isOpen } onClick={ props.onOpen }>
+						<AnalysisHeaderIcon icon={ props.isOpen ? angleUp : angleDown } color={ colors.$color_grey_dark } size="20px" />
+						<AnalysisTitle> { props.title + " (" + props.children.length + ")" } </AnalysisTitle>
+					</AnalysisHeaderButton>
+				</h3>
+			</div>
+			<ul >
```
afercia: Please remove the space after ul.

Xyfi:
I pushed some changes to wrap the component in a separate class based component to handle de collapsing. Still needs some work.

afercia:
Today I've learned: empty lists e.g. `<ul></ul>` are valid in HTML5 (this was not the case in HMTL 4.0.1):
https://www.w3.org/TR/html51/grouping-content.html#the-ul-element
> Content model:
> 	Zero or more li and script-supporting elements.

composites/Plugin/ContentAnalysis/components/AnalysisCollapsible.js
```diff
@@ -0,0 +1,170 @@
...
+/**
+ * A collapsible header used to show sets of analysis results. Expects list items as children.
+ *
+ * @param {object} props The properties for the component.
+ * @returns {ReactElement} A collasible analysisresult set.
+ */
+export const AnalysisCollapsibleStateless = ( props ) => {
+	return (
+		<AnalysisHeaderContainer>
+			<div>
```
afercia: Do we really need this div?

afercia:
CR done, one minor consideration.

I have a doubt about the props: it `AnalysisCollapsibleStateless` is going to be used exclusively as base for the statefull one, shouldn't the _required_ props be the same on both components?

afercia:
CR considerations addressed in the latest commit 👍

afercia:
The SVG icons introduced here should also use
`role="img" aria-hidden="true" focusable="false"`
see per https://github.com/Yoast/yoast-components/issues/255. I can add that, if no objections.

afercia:
CR 2 done 🚧 

- the button now contains a H3: that's invalid HTML. Note: collapsible headers in My Yoast are just buttons that contain a span, no headings
- buttons not intended to submit a form should always have a `type="button"` attribute. if accidentally placed within a form, this button will submit the form in some browsers that assume the default type is "submit" 
- the icon is not vertically aligned with the text: since now the component is a flex container, alignments of children should rely on flex and some properties should be removed; note: this button is very very similar to the `IconButton` component, I'd consider to extend from there because that fixes a few browsers bug
- not sure why the flex container is using `justify-content: space-between;` when then the text is `flex: 1 1 auto;`:  I guess we don't want `space-between`

About the SVG role, focusable... done in #280.

afercia:
First pass to refactor in order to use `IconButton`. Seems to me can work nicey this way, we just need to make a decision about which styles we want. 

Please test with this content in ContentAnalysis.js:

```
import React from "react";
import AnalysisCollapsible from "../../ContentAnalysis/components/AnalysisCollapsible";
import { IconButton } from "../../Shared/components/Button";

const SnippetPreviewDiv = styled.div`
	height: 700px;
	width: 80%;
	margin: 0 auto;
	background-color: white;

	button{
		margin-right: 10px;
	}

	.with-max-width {
		max-width: 152px;
	}
`;


export default class SnippetPreview extends React.Component {
	/**
	 * The constructor.
	 *
	 * @constructor
	 *
	 * @param {Object} props The props to use.
	 */
	constructor( props ) {
		super( props );
	}

	render() {
		return (
			<SnippetPreviewDiv>
				<h2>Buttons</h2>

				<h2> Analysis results </h2>
				<AnalysisCollapsible title="Problems">
					<li>Dont worry.</li>
					<li>About a thing.</li>
					<li>Cause every little thing,</li>
					<li>Is gonna be alright.</li>
				</AnalysisCollapsible>
			</SnippetPreviewDiv>
		);
	}
}
```

afercia:
Maybe consider to add an unstyled (or with very basic style) "base icon button" to be re-used acrosso components.

Xyfi:
Although I like the idea of reusing components (duh), focus styles are not desirable for the header.

title

afercia:
CR done 👍  just 2 small things I've fixed in the last commit. Most notably, the `IconButton` already has `type="button"` by default so no need to add it again.

afercia:
When there are no children (i.e. no analysis problems)  the check for `props.children.length` will error. What the component should do when there are no problems? Not being rendered at all? In this case we can just `return null` early. Or should the component display "Problems (0)" or anything else?

Additionally, when there's only 1 problem, `children` is not an array. It it a node instead, so it doesn't have a length:

screen shot 2017-09-21 at 10 18 10

composites/Plugin/ContentAnalysis/components/AnalysisCollapsible.js
```diff
@@ -0,0 +1,163 @@
...
+export const AnalysisCollapsibleStateless = ( props ) => {
+	return (
+		<AnalysisHeaderContainer>
+			<AnalysisHeaderButton
+				aria-expanded={ props.isOpen }
+				onClick={ props.onToggle }
+				icon={ props.isOpen ? angleUp : angleDown }
+				iconColor={ colors.$color_grey_dark }
+			>
+				<AnalysisTitle>{ props.title + " (" + props.children.length + ")" }</AnalysisTitle>
```
afercia: We shouldn't assume there are always children with a length:
- when there are no children, this will error because `children` is undefined: `TypeError: Cannot read property 'length' of undefined`
- when there's just 1 problem, `children` is not an array: it's a node (react.element) so it doesn't have a length property

afercia:
Acceptance not OK.
The component should better handle when there are no children or just one children.

Xyfi:
Fixed the handling of the number of children. Use the example in the top comment to test all possible situations (no children, 1 child and multiple children).

utils/reactUtils.js
```diff
@@ -0,0 +1,18 @@
+/**
+ * Determines the child count based on the children prop.
+ *
+ * @param {ReactElement|ReactElement[]} children Component children prop.
+ *
+ * @returns {number} Number of children
+ */
+export function getChildCount( children ) {
+	let count = 0;
+	if( children ) {
```
afercia: missing space after if

utils/reactUtils.js
```diff
@@ -0,0 +1,18 @@
+/**
+ * Determines the child count based on the children prop.
+ *
+ * @param {ReactElement|ReactElement[]} children Component children prop.
+ *
+ * @returns {number} Number of children
+ */
+export function getChildCount( children ) {
```
afercia: I'd consider to rename it to `getChildrenCount`

afercia:
CR done
just needs a very few minor CS in `reactUtils.js`

afercia:
CR 👍 
Removed a couple spaces to avoid React comments in the output.

IreneStr:
Acceptance 👌 

One thing to keep in mind when implementing: The count is based on direct children. Therefore, it's not possible (in the current implementation) to use a Table as the 'content' of the header (because the Table is the header's child in that case, not the li's inside the table). 

schermafbeelding 2017-09-27 om 11 27 14

PR #241, opened 2017-08-23
Implement icon radio button

### LINKED ISSUE #230: Implement IconRadioButton component
Implement an IconRadioButton component, based on the IconButton.

- `<input type="radio">`.
- The icon and background color should change when the button is clicked.
- The IconRadioButton should have a toggle state.

Depends on #190

IreneStr:
Implements the ChangingIconButtonComponent. 

To test, copy the following to ContentAnalysis.js [updated]:

````js
import React from "react";

import { eye } from "./../../../../style-guide/svg";
import IconButtonToggle from "../../Shared/components/IconButtonToggle.js";

	min-height: 700px;
	width: 100%;
	background-color: white;
`;

/**
 *
 */
export default class ContentAnalysis extends React.Component {
	/**
	 * Constructor for the component.
	 *
	 * @param {Object} props The props to assign to the current component.
	 *
	 * @constructor
	 */
	constructor( props ) {
		super( props );

		this.state = {
			pressed: "",
		};

		this.onRadioButtonClick = this.onRadioButtonClick.bind( this );
	}

	/**
	 * Handles the onButtonClick event.
	 *
	 * @param {string} id The radio button id.
	 * @returns {void}
	 */
	onRadioButtonClick( id ) {
		return this.state.pressed === id ? this.setState( { pressed: "" } ) : this.setState( { pressed: id } );
	}

	/**
	 * Renders a single radio button.
	 *
	 * @param {string} id The radio button id.
	 * @returns {void}
	 */
	renderRadioButton( id ) {
		return <ChangingIconButton
			onClick={ () => this.onRadioButtonClick( id ) }
			id={ id }
			icon={ eye }
			pressed={ id === this.state.pressed }
		/>;
	}

	render() {
		return ( <ContentAnalysisContainer>
			<br />
			{ this.renderRadioButton( "ChangingIconButton" ) }
			{ this.renderRadioButton( "RadioButton2" ) }
			{ this.renderRadioButton( "RadioButton3" ) }
			{ this.renderRadioButton( "RadioButton4" ) }
			{ this.renderRadioButton( "RadioButton5" ) }
			{ this.renderRadioButton( "RadioButton6" ) }
			{ this.renderRadioButton( "RadioButton7" ) }

		</ContentAnalysisContainer>
		);
	}
}

````
## Testing
The mockups from the issue are outdated. 

New sketch:
knops
Eye icon should stay on hover and pressed state. Border should disappear on hover.

### Technical choices 
Because of the way the button will be implemented using React and Redux, using a radio button instead of a regular button had no real advantages (despite the fact that a maximum of one button should be active at any given time). Also, it can be argued that these buttons act more like a toggle than a radio button. 

In addition, too many properties of the IconButton should be altered to make it useful to use that component as a basis for this IconButtonToggle component, the maxHeight property in combination with the IE bug being the most problematic. 

Fixes #230

### CODE REVIEW

**Comment by boblinthorst** [comment]
CR: the radio buttons don't seem to have an onClick method passed. And they don't take care of the clicking themselves, so I'm not sure how this is supposed to function :(,

At @Xyfi 's suggestion, the test content has been altered in my tests, which is how it's probably going to work in the future.

```
import React from "react";

import IconRadioButton from "./../../Shared/components/RadioButton.js";
import eye from "./../../../../style-guide/svg/eye.svg";
import eyeSlash from "./../../../../style-guide/svg/eye-slash.svg";

	min-height: 700px;
	width: 100%;
	background-color: white;
`;

/**
 *
 */
export default class ContentAnalysis extends React.Component {

	constructor( props ) {
		super( props );

		this.state = {
			checked: "RadioButton"
		}
	}

	renderRadioButton( id, group ) {
		return <IconRadioButton onClick={ () => this.setState( { checked: id } ) } key={ id } name={ group } id={ id } checkedIcon={ eyeSlash } uncheckedIcon={ eye } checked={ id === this.state.checked }/>;
	}

	render() {
		return ( <ContentAnalysisContainer>
				<br />
				{ this.renderRadioButton( "RadioButton", "group1" ) }
				{ this.renderRadioButton( "RadioButton2", "group1" ) }
				{ this.renderRadioButton( "RadioButton3", "group1" ) }
				{ this.renderRadioButton( "RadioButton4", "group1" ) }
				{ this.renderRadioButton( "RadioButton5", "group1" ) }
				{ this.renderRadioButton( "RadioButton6", "group1" ) }
				{ this.renderRadioButton( "RadioButton7", "group1" ) }

			</ContentAnalysisContainer>
		);
	}
}

```

afercia:
Sorry to chime in, as I may have missed some internal discussion. I see some issues here, for example:
- inputs with `display: none;` won't be announced even if they have a label because well... they're hidden
- wondering why we're using a `<label>` element with an `aria-label="RadioButton"` attribute when the aria-label could be applied directly on the input
- the `aria-checked` attribute shouldn't be set on the label, but on the input instead, as the checkedness is a property of the input

Anyways, before these considerations, I'm not sure what the goal of this component is and why we're using a radio button in the first place. If the goal is to have a button that has a "pressed" state, why simply not use a button?

Worth also considering radio buttons are meant to allow a single choice in a set of choices that answer the same question, so I'm not sure they're appropriate for this use case.

afercia:
CR done 🚧 

Some points to address:

This component is not changing icon any longer, but it's still named `ChangingIconButton` (and `ChangingIconButtonBase`): I'd consider to rename it

Note: The `eye-slash` icon is unused at the moment, it doesn't harm anything so I guess it's fine to keep it.

`export const ChangingIconButtonBase`
I guess there's no need to export the base button, it's used just internally and can't even be used standalone because some props are only passed by the wrapper button.

Firefox padding: FF applies some default padding from the user agent stylesheet that causes the icon to not be centered: please add `padding: 0`:
screen shot 2017-10-16 at 09 26 27

a11y:
the button needs a label with a meaningful text: we can't use the `id` because it serves a different purpose (used in the state to match the pressed state). We should add an `ariaLabel` required prop, and I guess it should be passed together with the `id`.

The aria-label should be translatable, that means it should use `react-intl` (also the tests should be updated accordingly).

a11y:
at the moment there's no focus style at all: focus indication should always be provided, in absence of a specific design for that, I'd suggest to just remove `outline: none;` and use the browsers default

Tests:
the two tests have the same name "the unpressed ChangingIconButton matches the snapshot"

Xyfi:
CR :+1:

Xyfi:
Acceptance :+1:

Issue #198, opened 2017-08-03
Reactify Help Center - KB tab


hedgefield:
The KB page functions largely the same as it does now. I've marked any changes I've made in bold, the rest can pretty much be copied one-to-one.

![help kb search](https://user-images.githubusercontent.com/6073772/28928150-c9884d2e-786c-11e7-9e47-5dc494c07779.png)

![help kb](https://user-images.githubusercontent.com/6073772/28928102-abd39946-786c-11e7-8329-d83233cbd965.png)

![help center kb article](https://user-images.githubusercontent.com/6073772/29705842-0dda7d84-8980-11e7-825b-13fede5a3bd4.png)

- When you open the page, it displays a search bar with **an icon and a button. The placeholder reads `Type here to search the Yoast Knowledge Base`.**
- When you type something, you have to hit enter (or the button) to submit. No live search or autocomplete (yet).
- The entire paper can scale vertically depending on how many results need to be shown.
- **The results are shown in a zebra stripe list.**
- Navigating away to another tab or closing the help center clears the search.

IreneStr:
Fixed in https://github.com/Yoast/wordpress-seo/pull/7948

PR #243, opened 2017-08-24
Implement AlgoliaSearcher

### LINKED ISSUE #213: Implement Algolia Searcher
- Implement changes in AlgoliaSearcher.js.
- Move code from SearchResult.js to SearchResults.js.
- Implement SearchResultDetail component (change incorrect comments!).
- Implement SearchResults component.
- Implement name changes in Loading.js.
- Implement SearchBar as a styled component.
- Add search.svg
- Add `react-intl` and `svg-url-loader` dependencies. (+ Look into svg-react-loader instead (ask @Xyfi)) 
- Add `svg-url-loader` (or replacement) to webpack.config.js

Part 3 of the implementation of pull #203

IreneStr:
### To do
- Make texts passed as props translatable using react intl. See https://github.com/yahoo/react-intl/issues/438
- Style buttons (@hedgefield will put a screenshot below) and put them at the right place (use current help center as reference).
- I'm not sure where the warnings about `Accessing PropTypes via the main React package is deprecated` are coming from.

### Testing

Copy the following to ContentAnalysis.js:
````js
import React from "react";

import AlgoliaSearcher from "../../../AlgoliaSearch/AlgoliaSearcher";

	min-height: 700px;
	width: 100%;
	background-color: white;
`;

/**
 *
 */
		<AlgoliaSearcher />
	</ContentAnalysisContainer>;
}
````
Fixed #213

### CODE REVIEW

**Comment by hedgefield** [comment]
Here is the styling of the buttons. I changed `Open` to `View in KB` to make it a bit more descriptive.

One other thing: in this mockup I removed the purple AMP KB header. It takes up a lot of unneccessary space, especially since we can't use that space for the buttons (those would have to be above it, outside the iframe). I discussed with @diedexx that it is possible to ignore the header when doing the iframe request, so please add that to this PR if possible.

help center kb article

**Comment by diedexx** [comment]
The kbsearch sends a GET parameter (`wpseo-kb-search`) with each request telling it comes from the plugin. The same can be done here. 
We then need to remove the purple banner on the AMP version of the website from the page if the GET param was set. https://github.com/Yoast/yoast.com-ms/blob/master/site/web/app/plugins/clean-amp/clean-amp.php#L16

**Comment by Xyfi** [comment]
The `back` and `go to KB` button redesign has been moved to three separate issues: #251, #252, #253.

File: composites/AlgoliaSearch/AlgoliaSearcher.js [hunk 1]
```diff
@@ -108,20 +134,18 @@ class AlgoliaSearcher extends React.Component {
...
 	 *
 	 * @returns {void}
 	 */
 	updateSearchResults() {
-		this.setState( { searching: true } );
-
-		this.performSearch( this.state.searchString ).then( this.processResults.bind( this ) ).catch( this.processSearchError.bind( this ) );
+		this.performSearch( this.state.searchString )
+		    .then( this.processResults.bind( this ) )
+		    .catch( this.processSearchError.bind( this ) );
```
afercia: Minor, but alignments shouldn't use mixed tab characters and spaces.

composites/AlgoliaSearch/AlgoliaSearcher.js
```diff
@@ -108,20 +134,18 @@ class AlgoliaSearcher extends React.Component {
...
 	}
 
 	/**
-	 * Performs a search with a given search string on the algolia index which
-	 * information was passed in the AlgoliaSearcher's props.
+	 * Performs a search with the given search string withing the Algolia index.
```
afercia: Typo: withing

composites/AlgoliaSearch/AlgoliaSearcher.js
```diff
@@ -141,182 +165,146 @@ class AlgoliaSearcher extends React.Component {
...
 
 	/**
-	 * Maps the results to SearchResult components.
+	 * Log any occuring error and render a search error warning.
 	 *
-	 * @param {object} results The results returned by Algolia.
+	 * @param {string} errorMessage The message to display.
 	 *
-	 * @returns {Array} Array containing the mapped search results.
+	 * @returns {ReactElement} A p tag with a warning that the search was not completed.
```
afercia: Technically, a "tag" is what you use to in the markup language. This return a p element instead.

composites/AlgoliaSearch/AlgoliaSearcher.js
```diff
@@ -141,182 +165,146 @@ class AlgoliaSearcher extends React.Component {
...
+	renderError() {
+		const errorMessage = this.props.intl.formatMessage( messages.errorMessage );
 
+		a11ySpeak( errorMessage );
+
+		return (
+			<p>{ errorMessage }</p>
+		);
+	}
 	/**
```
afercia: missing empty line

composites/AlgoliaSearch/AlgoliaSearcher.js
```diff
@@ -141,182 +165,146 @@ class AlgoliaSearcher extends React.Component {
...
-		// We'll check to see whether no results are returned.
-		if ( resultsCount <= 0 && this.state.searchString !== "" ) {
-			return this.renderNoResultsFound();
-		}
-
-		if ( resultsCount === 0 ) {
-			return searchResultContent;
+	createSearchBar() {
+		if ( this.state.currentDetailViewIndex !== -1 ) {
+			return "";
```
afercia: shouldn't this return `null` instead? If so, please update also the docblock.

composites/AlgoliaSearch/AlgoliaSearcher.js
```diff
@@ -141,182 +165,146 @@ class AlgoliaSearcher extends React.Component {
...
-					   aria-label={ this.props.openLabel }
-					   target="_blank">
-						{ this.props.open }
-					</a>
-				</div>
-
-				<ArticleContent post={post} iframeTitle={this.props.iframeTitle}/>
-			</div>
-		);
+		return "";
```
afercia: shouldn't this return null instead? If so, please update also the docblock.

composites/AlgoliaSearch/AlgoliaSearcher.js
```diff
@@ -141,182 +165,146 @@ class AlgoliaSearcher extends React.Component {
...
-		a11ySpeak( this.props.errorMessage );
+	getLoadingIndicator() {
+		const loadingPlaceholder = this.props.intl.formatMessage( messages.loadingPlaceholder );
 
-		return ( <p>{ this.props.errorMessage }</p> );
+		if ( this.state.searching ) {
+			return <Loading placeholder={ loadingPlaceholder } />;
+		}
+
+		return "";
```
afercia: shouldn't this return null instead? If so, please update also the docblock.

composites/AlgoliaSearch/AlgoliaSearcher.js
```diff
@@ -141,182 +165,146 @@ class AlgoliaSearcher extends React.Component {
...
 		}
 
-		return <div>
-			{ searchBar }
-			{ errorMessage }
-			{ loadingIndicator }
-			{ resultsHeading }
-			{ results }
-		</div>;
+		return "";
```
afercia: shouldn't this return null instead?

composites/AlgoliaSearch/AlgoliaSearcher.js
```diff
@@ -336,48 +332,29 @@ class AlgoliaSearcher extends React.Component {
 	 * @returns {ReactElement|string} Returns a header if there are search results. Otherwise returns an empty string.
 	 */
 	determineResultsHeading() {
-		if ( this.state.results.length === 0 ) {
+		if ( this.state.currentDetailViewIndex === -1 || this.state.results.length === 0 ) {
 			return "";
```
afercia: shouldn't this return null instead? If so, please update also the docblock.

composites/AlgoliaSearch/AlgoliaSearcher.js
```diff
@@ -336,48 +332,29 @@ class AlgoliaSearcher extends React.Component {
...
+	intl: intlShape.isRequired,
 };
 
 AlgoliaSearcher.defaultProps = {
-	foundResultsText: "Number of search results: %d",
-	noResultsText: "No results found.",
-	headingText: "Search the Yoast knowledge base",
-	searchButtonText: "Search",
-	searchResultsHeading: "Search results",
 	iframeTitle: "Knowledge base article",
```
afercia: This should be translatable.

composites/AlgoliaSearch/Loading.js
```diff
@@ -9,11 +9,11 @@ import PropTypes from "prop-types";
  * @constructor
```
afercia: it this @constructor notation correct?

composites/AlgoliaSearch/SearchBar.js
```diff
@@ -1,47 +1,168 @@
...
+const SearchLabel = styled.label`
+	width: 2em;
+	height: 2em;
+	float: left;
+	margin-top: 0.25em;
+`;
+
+SearchLabel.propTypes = {
+	className: PropTypes.string,
+	htmlFor: PropTypes.string,
```
afercia: not sure htmlFor should be specified as prop

composites/AlgoliaSearch/SearchBar.js
```diff
@@ -1,47 +1,168 @@
...
 };
 
+SearchLabel.defaultProps = {
+	className: "",
+	htmlFor: "",
```
afercia: same as above (and `for` attributes can't be empty)

composites/AlgoliaSearch/SearchBar.js
```diff
@@ -1,47 +1,168 @@
...
+
+	/**
+	 * Defines the debounced doFormSubmission function to be triggered after the user quits typing for 1 second.
+	 *
+	 * @returns {void}
+	 */
+	componentWillMount() {
+		this.doFormSubmission = debounce( ( searchString ) => {
+			this.props.submitAction( searchString );
+		}, 1000 );
```
afercia: I think this debounced call should be canceled also when the component unmounts. Right now, it gets canceled only when submitting, see `this.doFormSubmission.cancel();`

composites/AlgoliaSearch/SearchResultDetail.js
```diff
@@ -0,0 +1,95 @@
...
+const OpenLink = styled.a`
+	float: right;
+	display: inline-block;
+	text-decoration: none;
+	font-size: 13px;
+	line-height: 26px;
+	height: 28px;
+	margin: 0;
+	padding: 0 10px 1px;
+	cursor: pointer;
```
afercia: liks have cursor pointer by default, consider to remove this

composites/AlgoliaSearch/SearchResults.js
```diff
@@ -0,0 +1,202 @@
...
+import colors from "../../style-guide/colors.json";
+
+const messages = defineMessages( {
+	noResultsText: {
+		id: "searchresults.noresultstext",
+		defaultMessage: "No results found.",
+	},
+	foundResultsText: {
+		id: "searchresults.foundresultstext",
+		defaultMessage: "Found {resultsCount} { resultsCount, plural, one {result} other {results} }",
```
afercia: for better translations, this should use placeholders and replacements. Or, to simplify, a text like:
`Number of results found: %d` is considered acceptable even for languages that have multiple plural forms

composites/AlgoliaSearch/SearchResults.js
```diff
@@ -0,0 +1,202 @@
...
+
+/**
+ * The link to the search result.
+ */
+const SearchResultLink = styled.a`
+	color: ${ colors.$color_black };
+	padding: 10px;
+	
+	&:hover, &:focus {
+		color: ${ colors.$color_pink_dark };
```
afercia: for a11y, a focus indication with color alone is not sufficient. For now, this keeps the native outline styles so it's OK. But, if the outline is going to be reset in the implementation, then we'd need to add something else for focus.

composites/AlgoliaSearch/SearchResults.js
```diff
@@ -0,0 +1,202 @@
...
+	post: PropTypes.object.isRequired,
+};
+
+/**
+ * Wraps the search results.
+ */
+const SearchResultsWrapper = styled.div`
+	margin-top: 20px;
+	width: 100%;
+	display: block;
```
afercia: div elements are natively display: block and full width, consider to remove these two properties.

composites/AlgoliaSearch/SearchResults.js
```diff
@@ -0,0 +1,202 @@
...
+`;
+
+/**
+ * Create the ReactElement to render a list of search results.
+ *
+ * @param {object} props The React props.
+ *
+ * @returns {ReactElement} A list of search results.
+ *
+ * @constructor
```
afercia: is this notation correct> is the whole class a constructor or...

composites/AlgoliaSearch/SearchResults.js
```diff
@@ -0,0 +1,202 @@
...
+class SearchResults extends React.Component {
+
+	/**
+	 * Constructs the component and sets its initial state.
+	 *
+	 * @param {Object} props The props to use for this component.
+	 *
+	 * @constructor
+	 */
+	constructor( props ) {
```
afercia: is it this the constructor?

afercia:
CR done 🚧 
A few points to address.

Additionally, I've noticed a couple things. I understand the styling is temporary and will probably need to be adjusted during implementation, however worth considering that:
- when searching, the "loading" placeholder appears and then the results but the content has a weird "jump" effect
- in Firefox, the search term entered in the field is not visible... are we testing in all browsers, right? 🙂 

Firefox (entered term: "cornerstone"):

screen shot 2017-08-31 at 15 22 55

composites/AlgoliaSearch/SearchBar.js
Xyfi: Removed them completely. className also is a react specific property.

composites/AlgoliaSearch/SearchResults.js
Xyfi: Shouldn't we use the tools that `react-intl` provides us?

afercia:
CR 🚧 

Can't actually test.

This PR now introduces a modified `yarn.lock` file with a reference to `react-dock@^0.2.4` however, no `react-dock` is specified in `package.json` so I get an error `Module build failed` and yarn doesn't solve it because it doesn't know it has to install something new.

Also:
this PR introduces `react-intl` for translations but in other components we're using Jed. Noting just to be sure this has been discussed internally and it's OK.

afercia:
Hm after a few more `yarn` build is now successful. Concern about the missing entry in package.json stands though.

afercia:
Not sure why searching for a term displays also the "No results found." For example, when I search for `development server` I get 2 results and below I see also "No results found.":

Dumping `resultsCount` inside the render function, seems render runs twice, first time 0 results and second time 2 results:

screen shot 2017-09-08 at 13 25 29


Another important point is the lack of a "submit search" button. We've had an interesting discussion about the Knowledge Base for My Yoast and decided to re-introduce the "submit search" button, see 
https://github.com/Yoast/my-yoast/issues/689#issuecomment-318695998
https://github.com/Yoast/yoast-components/issues/198

for accessibility and consistency, I'd say there should be a button here too. /cc @hedgefield

afercia:
CR 2 done 🚧 

The "No results found." issue seems something should be investigated to me.
Also, needs a design decision about the "submit search" button (sorry for not having pointed out that earlier).

afercia:
OK I think some of the misunderstanding here is because of the code to be used for testing, it uses two SearchResults components so onload there are 2 searches. Not sure about some of the passed props too. I'd need some more details.

composites/AlgoliaSearch/AlgoliaSearcher.js
```diff
@@ -1,16 +1,42 @@
...
+	},
+	errorMessage: {
+		id: "algoliasearcher.errormessage",
+		defaultMessage: "Something went wrong. Please try again later.",
+	},
+	searchResultsHeading: {
+		id: "algoliaseacher.searchresultsheading",
+		defaultMessage: "Search results",
+	},
+	iframeTitle: {
```
afercia: this is not used, see duplicate in SearchResultDetail.js

composites/AlgoliaSearch/AlgoliaSearcher.js
```diff
@@ -336,48 +336,27 @@ class AlgoliaSearcher extends React.Component {
 	 * @returns {ReactElement|string} Returns a header if there are search results. Otherwise returns an empty string.
```
afercia: technically, "Otherwise returns an empty string" is not correct, because it returns `null`.

composites/AlgoliaSearch/AlgoliaSearcher.js
```diff
@@ -336,48 +336,27 @@ class AlgoliaSearcher extends React.Component {
...
 	 */
 	determineResultsHeading() {
-		if ( this.state.results.length === 0 ) {
-			return "";
+		if ( this.state.currentDetailViewIndex === -1 || this.state.results.length === 0 ) {
+			return null;
 		}
 
-		return <h2 className="screen-reader-text">{this.props.searchResultsHeading}</h2>;
+		return <h2 className="screen-reader-text">
```
afercia: I can't see this heading appearing after a search, maybe the logic and conditional should be checked a bit.

composites/AlgoliaSearch/AlgoliaSearcher.js
```diff
@@ -141,182 +168,147 @@ class AlgoliaSearcher extends React.Component {
...
+			<p>{ errorMessage }</p>
+		);
 	}
 
 	/**
-	 * Renders the navigation buttons and the article content.
+	 * Creates the Search Bar component.
 	 *
-	 * @returns {ReactElement} A div with navigation buttons and the article content.
+	 * @returns {ReactElement|string} Searchbar component. Returns empty string if we're in the detail view.
```
afercia: technically, "Returns empty string if ..." is not correct, because it returns `null`.

composites/AlgoliaSearch/AlgoliaSearcher.js
```diff
@@ -51,29 +73,38 @@ class AlgoliaSearcher extends React.Component {
...
 	 *
-	 * @param {object} clickEvent The React SyntheticEvent.
+	 * @param {string} searchString The entered search string.
 	 *
 	 * @returns {void}
 	 */
-	searchButtonClicked( clickEvent ) {
-		let searchString = clickEvent.target.getElementsByTagName( "input" )[ 0 ].value;
-
+	onSearchButtonClick( searchString ) {
```
afercia: pending decision on the button, see #285, there's no button right now. The function name should reflect what the function does.

composites/AlgoliaSearch/AlgoliaSearcher.js
```diff
@@ -141,182 +168,147 @@ class AlgoliaSearcher extends React.Component {
...
 	 */
-	showDetail( resultIndex ) {
+	addResultToUsedQueries( currentDetailViewIndex ) {
+		let post = this.state.results[ currentDetailViewIndex ];
 		let usedQueries = this.state.usedQueries;
-		let post = this.state.results[ resultIndex ];
 
 		usedQueries[ this.state.searchString ][ post.objectID ] = {
-			title: post.post_title,
+			title: post.postTitle,
```
afercia: ESLint disagrees but this should really be `post_title`. The returned  `post` object doesn't have a `postTitle ` property, it does have a `post_title ` one. Maybe disable the ESLint rule for this line?

Note, the post title saved in the used queries is actually `undefined` because of this:

screen shot 2017-09-18 at 14 45 36

composites/AlgoliaSearch/AlgoliaSearcher.js
```diff
@@ -108,20 +137,18 @@ class AlgoliaSearcher extends React.Component {
...
 	 *
 	 * @returns {void}
 	 */
 	updateSearchResults() {
-		this.setState( { searching: true } );
-
-		this.performSearch( this.state.searchString ).then( this.processResults.bind( this ) ).catch( this.processSearchError.bind( this ) );
+		this.performSearch( this.state.searchString )
+		    .then( this.processResults.bind( this ) )
+		    .catch( this.processSearchError.bind( this ) );
```
afercia: there are still mixed tab characters and spaces in these 2 lines

composites/AlgoliaSearch/AlgoliaSearcher.js
```diff
@@ -141,182 +168,147 @@ class AlgoliaSearcher extends React.Component {
...
-			loadingIndicator = <Loading loadingPlaceholder={ this.props.loadingPlaceholder } />;
+		return null;
+	}
+
+	/**
+	 * Determines what the search results view needs to look like.
+	 *
+	 * @returns {ReactElement|string} Returns a specific search result object based on state.
+	 */
+	determineSearchResultsView() {
```
afercia: in this function there are 3 if and a default return, not sure it there are other cases to take into account apart from `this.state.currentDetailViewIndex === -1` and `this.state.currentDetailViewIndex !== -1`. Mayne this could be a bit optimized?

composites/AlgoliaSearch/SearchBar.js
```diff
@@ -20,28 +51,110 @@ const onSubmit = ( submitEvent, props ) => {
  * @returns {ReactElement} A div with the searchbar.
  * @constructor
```
afercia: not sure the @constructor notation here is correct

composites/AlgoliaSearch/SearchBar.js
```diff
@@ -20,28 +51,110 @@ const onSubmit = ( submitEvent, props ) => {
...
+				<form onSubmit={ this.onSubmit.bind( this ) }>
+					<SearchLabel htmlFor="search-input">
+						<Icon icon={ SearchIcon } color="inherit" size="30px" />
+						<span className="screen-reader-text">
+							{ headingText }
+						</span>
+					</SearchLabel>
+					<SearchBarInput
+						onChange={ this.onSearchChange.bind( this ) }
+						onSubmit={ ( event ) => {
```
afercia: is this necessary? form submission is already prevented on the form itself, see `this.onSubmit()`

composites/AlgoliaSearch/SearchResultDetail.js
```diff
@@ -0,0 +1,98 @@
...
+
+const Detail = styled.section``;
+
+const Nav = styled.nav`
+	padding: 8px;
+`;
+
+const OpenLink = styled.a`
+	float: right;
+	display: inline-block;
```
afercia: floated elements behave like block elements (as if they had `display: block`) so `display: inline-block;` doesn't do anything here.

composites/AlgoliaSearch/SearchResultDetail.js
```diff
@@ -0,0 +1,98 @@
...
+	white-space: nowrap;
+	box-sizing: border-box;
+`;
+
+/**
+ * Create the JSX to render the SearchResultDetail component.
+ *
+ * @param {object} props The React props.
+ * @returns {ReactElement} A SearchResultDetail component.
+ * @constructor
```
afercia: not sure the @constructor notation is correct here

composites/AlgoliaSearch/SearchResults.js
```diff
@@ -0,0 +1,198 @@
+import React from "react";
+import PropTypes from "prop-types";
+import a11ySpeak from "a11y-speak";
+import { injectIntl, intlShape, defineMessages } from "react-intl";
+import styled from "styled-components";
+import { ZebrafiedListTable } from "../basic/Table/ListTable";
```
afercia: for consistency, please separate the absolute / relative imports

composites/AlgoliaSearch/SearchResults.js
```diff
@@ -0,0 +1,198 @@
...
+ * Returns a single search result item.
+ *
+ * @param {Object} props Properties of the SearchResult component.
+ * @returns {ReactElement} The rendered search result.
+ */
+export function SearchResult( props ) {
+	let post = props.post;
+
+	return (
+		<Row {...props}>
```
afercia: please add spaces within brackets

composites/AlgoliaSearch/SearchResults.js
```diff
@@ -0,0 +1,198 @@
...
+	 * Maps the results to SearchResult components.
+	 *
+	 * @param {Object} results The results returned by Algolia.
+	 *
+	 * @returns {Array} Array containing the mapped search results.
+	 */
+	resultsToSearchItem( results ) {
+		return results.map( ( result, index ) => {
+			return <SearchResult
+				height="32px"
```
afercia: this causes a `height` HTML attribute to be printed out on a `<li>` element, which is not valid HTML:
`Attribute height not allowed on element li at this point.`

I guess this should be investigated in the Table Row component?

composites/AlgoliaSearch/SearchResults.js
```diff
@@ -0,0 +1,198 @@
...
+					this.props.handler( index );
+				} }
+			/>;
+		} );
+	}
+
+	/**
+	 * Renders the search results list.
+	 *
+	 * @returns {ReactElement} A div with either the search results, or a div with a message that no results were found.
```
afercia: actually, when there are no results, a `p` is returned

afercia:
CR done 🚧 
Some points to address/check, please see the review.

When implementing relevant changes, please don't forget to update the testing code included in the PR. It took me a while to understand the testing code above is outdated and given the size of this PR things get a bit complicated to follow and test.

Some props in the testing code above are now in the state, the `SearchResultDetail` component shouldn't even be there, etc.

I wouldn't recommend to test 2 `<SearchResults>` at the same time, since the state gets updated 2 times and you will see some search results _and_ the "no results" message at the same time.

afercia:
Also, not sure the error message works correctly when used for Speak: seems to me it is immediatly followed by another message (no result) so it is not announced:

screen shot 2017-09-18 at 16 51 15

afercia:
Note: `currentDetailViewIndex` has a mixed use
- as the index of the post to view in the details view
- as a flag to distinguish the search results list view from the details one

it took me a while to understand this. Wouldn't be better to use a specific state property for the flag? `isDetailView` or similar? `currentDetailViewIndex` doesn't exactly tell what it does. At least, consider to use a better name?

composites/AlgoliaSearch/SearchResults.js
IreneStr: `Number of results found: %d` seems easier to me, and is consistent with My Yoast

composites/AlgoliaSearch/SearchResults.js
```diff
@@ -0,0 +1,205 @@
...
+ *
+ * @param {Object} props Properties of the SearchResult component.
+ * @returns {ReactElement} The rendered search result.
+ */
+export function SearchResult( props ) {
+	let post = props.post;
+
+	return (
+		<Row { ...props } >
+			<SearchResultLink href={ post.permalink } onClick={ props.handler }>
```
IreneStr: Rename `handler` to `onClick` for consistency?

composites/AlgoliaSearch/SearchResults.js
```diff
@@ -0,0 +1,205 @@
...
+
+/**
+ * The no results text paragraph.
+ */
+const NoResults = styled.p`
+	margin-left: 10px;
+`;
+
+/**
+ * Create the ReactElement to render a list of search results.
```
IreneStr: Creates the Search Results component.

composites/AlgoliaSearch/SearchResults.js
```diff
@@ -0,0 +1,205 @@
...
+					{ this.resultsToSearchItem( this.props.results ) }
+				</ZebrafiedListTable>
+			</SearchResultsWrapper>
+		);
+	}
+
+}
+
+SearchResults.propTypes = {
+	intl: intlShape,
```
IreneStr: .isRequired

composites/basic/Table/tests/RowTest.js
```diff
@@ -5,7 +5,7 @@ import { Row, RowResponsiveWrap } from "../../Table/Row";
 
 test( "the Row component matches the snapshot", () => {
 	const component = renderer.create(
-		<Row />
+		<Row heigth="48px"/>
```
IreneStr: rowHeight?

composites/AlgoliaSearch/SearchBar.js
```diff
@@ -20,28 +51,107 @@ const onSubmit = ( submitEvent, props ) => {
...
-	);
-};
+class SearchBar extends React.Component {
+
+	/**
+	 * Constructs the component and sets its initial state.
+	 *
+	 * @param {Object} props The props to use for this component.
+	 *
+	 * @constructor
```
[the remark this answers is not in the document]
Xyfi: Not necessary.

composites/AlgoliaSearch/AlgoliaSearcher.js
```diff
@@ -185,7 +185,7 @@ class AlgoliaSearcher extends React.Component {
 	}
 
 	/**
-	 * Sets all values required to display the detailed view of a search result.
+	 * Set the current view to the detail view and set the article index.
```
IreneStr: Sets

composites/AlgoliaSearch/AlgoliaSearcher.js
```diff
@@ -202,7 +202,7 @@ class AlgoliaSearcher extends React.Component {
 	}
 
 	/**
-	 * Resets the values associated with the displaying of the detailed search result view.
+	 * Set the current view to the search view.
```
IreneStr: Sets

IreneStr:
Acceptance 🚧 

The table is broken on mobile: fix height (+ remove padding on both sides?)
schermafbeelding 2017-09-21 om 09 46 43

afercia:
Maybe `height` should be `min-height`?

IreneStr:
I've fixed the above problems, except for the 'back to top' thingy on mobile.

@hedgefield Could you please share your opinion on the h2 placement (keep like this or move to the left to align with table start):
schermafbeelding 2017-09-21 om 11 09 05

and about the absence of the border in combination with the scrollbar:
schermafbeelding 2017-09-21 om 11 08 58



If you'd like to see changes, please create a new issue :)

Xyfi:
CR :+1: Made a minor change: Allowed maxWidth to be passed as a prop to the AlgoliaSearcher.

composites/AlgoliaSearch/SearchBar.js
```diff
@@ -1,47 +1,156 @@
...
+	 * @returns {void}
+	 */
+	componentWillUnmount() {
+		this.doFormSubmission.cancel();
+	}
+
+	/**
+	 * Handles the change event on the SearchBar.
+	 *
+	 * @param {DOMEvent} event The event being triggered on the SearchBar.
```
afercia: Technically, I guess this is not a DOM event, it's a React one (SyntheticEvent, which is a wrapper around the DOM event)

composites/AlgoliaSearch/SearchBar.js
```diff
@@ -1,47 +1,156 @@
...
+		this.setState( { searchString: event.target.value }, () => {
+			this.doFormSubmission( this.state.searchString );
+		} );
+	}
+
+	/**
+	 * Handles the search bar form submit.
+	 *
+	 * @param {DOMEvent} event The event being triggered on the SearchBar.
+	 *
```
afercia: Technically, I guess this is not a DOM event, see above.

afercia:
Using hardcoded IDs like `id="search-input"`, there's a chance to have conflicts with other IDs from other components, especially if the id value is so generic like `search-input`. We could consider to quickly improve this with a more specific id (`kb-search-input` ?) or consider to always use unique IDs, generated programmatically.

afercia:
There are a few, different, mockups around and probably that is a bit confusing. Perfectly understandable, however the current design (and functionality) doesn't match the requirements on the issue, see https://github.com/Yoast/yoast-components/issues/198

This is how it looks right now:

screen shot 2017-09-21 at 14 34 08

And this is how it should be:

28928150-c9884d2e-786c-11e7-9e47-5dc494c07779

- there is a search submit button
- quoting from the issue: "When you type something, you have to hit enter (or the button) to submit. No live search or autocomplete (yet)."
- there should be no visible text above the field
- the field should have a placeholder text (pending decision about how to properly use placeholders)
- the input field height should be 48px (based on the 8px grid) exactly like the button height

afercia:
Acceptance 👍 

Looks good as a first implementation. Will need adjustments, improvements, and to better meet the design provided in https://github.com/Yoast/yoast-components/issues/198. For that, we need to wait for other stuff, for example #284

hedgefield:
Judging by the last screenshot @IreneStr posted above, there are indeed some design issues. The back / view in kb buttons seem randomly styled? One looks like a WP button and one is flat with a blue outline? Is that still the case? Why aren't they the My Yoast style buttons like in the mockup?

Also, I discussed with @diedexx that it was possible to strip the purple yoast kb header from the AMP article, having that header in the help center is not necessary.

The designs in #198 are the latest ones.

afercia:
> Why aren't they the My Yoast style buttons like in the mockup?

Because the buttons (and link button) were not merged yet when this was merged 🙂 

> and to better meet the design provided in #198. For that, we need to wait for other stuff, for example #284

hedgefield:
Ahh well great :)

PR #256, opened 2017-08-31
Updated Onboarding Wizard header

### LINKED ISSUE #248: Onboarding Wizard Change top logo
Change the Yoast logo at the top of the configuration wizard into the Yoast SEO icon + the text `Yoast SEO for WordPress installation wizard`. Set it as an `<h1>`. It'll appear a little bigger than the mockup but that's fine.

screenshot_22_08_2017__16_11

Xyfi:
## Summary

* Updated onboarding wizard header

## Test instructions

* [edit by andrea]: run `grunt build` to compile the CSS
* Run the test environment and see the changes in the wizard.

Fixes #248


composites/OnboardingWizard/Header.js
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+
+import YoastSeoIcon from "../basic/YoastSeoIcon";
```
afercia: Shouldn't this have some docblock?

composites/OnboardingWizard/Header.js
```diff
@@ -0,0 +1,14 @@
...
+
+const Header = () => {
+	return (
+		<header className="yoast-wizard--header">
+			<YoastSeoIcon height={56} width={56}/>
```
afercia: please add spaces within brackets

composites/OnboardingWizard/Header.js
```diff
@@ -0,0 +1,14 @@
...
+			<h1>Yoast SEO for WordPress installation wizard</h1>
```
afercia: this should be translatable, except "Yoast SEO for WordPress" (we never translate plugin names for branding reasons)

composites/OnboardingWizard/onboarding-wizard.scss
```diff
@@ -44,6 +44,23 @@ body {
   box-sizing: border-box;
   width: 100%;
 
+  &--header {
+    text-align: center;
+
+    h1 {
+      color: $color_pink_dark;
+      margin: 0 0 -16px;
+      font-size: 1.2em;
```
afercia: this translates to a computed value of 19.2px, ideally we should use a value that computes to a number without decimals

composites/OnboardingWizard/onboarding-wizard.scss
```diff
@@ -44,6 +44,23 @@ body {
...
+      letter-spacing: 0.03em;
+      line-height: 2.6rem;
```
afercia: same as above (41.6px)

composites/OnboardingWizard/onboarding-wizard.scss
```diff
@@ -44,6 +44,23 @@ body {
...
+      font-weight: 100;
```
afercia: using a number as value for font-weight makes sense only when there's an actual font with that weight. The actual value computed here is "normal".

composites/OnboardingWizard/onboarding-wizard.scss
```diff
@@ -44,6 +44,23 @@ body {
...
+      @media screen and (max-width: 768px) {
+        font-size: 1.4em;
```
afercia: same as above, this translates to 22.4px

composites/OnboardingWizard/onboarding-wizard.scss
```diff
@@ -44,6 +44,23 @@ body {
...
   width: 100%;
 
+  &--header {
+    text-align: center;
+
+    h1 {
+      color: $color_pink_dark;
+      margin: 0 0 -16px;
+      font-size: 1.2em;
+      letter-spacing: 0.03em;
```
afercia: computes to 0.576px but I guess we can live with this?
Worth noting the actual rendering of this font in WordPress will be slightly different because here in the test environment the font is Arial, while WordPress uses the system fonts stack:
`-apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Oxygen-Sans, Ubuntu, Cantarell, "Helvetica Neue", sans-serif`

afercia:
CR done 🚧  with a few points to address.

Some considerations:
- I'm not sure about the new text being a H1. The actual "main content" in each "page" is the step title and maybe that should use a H1. Instead, the text under the logo seems to me it's a similar case like a "site title" in a post page, where the site title should be just text and the post title a H1
- the whole file `onboarding-wizard.scss` uses spaces for indentation 🙂  while not related to this PR, it would be great to convert everything to tab characters (I can volunteer for that if needed)
- maybe worth considering to add some left/right padding to the new text, as at some viewport width it will touch the screen edges (see screenshot below)

screen shot 2017-09-01 at 10 14 45

afercia:
Please add a `role="banner"` to the `<header>` element, see #258 and #260 

Note: the HTML validator will give a warning since the role is redundant. However, for the reasons explained in #258 the role is necessary. We could maybe consider to make the header a `<div role="banner" ...`

moorscode:
Please unblock https://github.com/Yoast/wordpress-seo/pull/7808 after merging!

afercia:
CR done 👍 
Discussed with @Xyfi how to make the header title translatable and implemented some changes.
The header title is now an optional prop of the `Header` component. If it's not passed, only the logo gets rendered.

herregroen:
Acceptance done 👍

PR #262, opened 2017-09-01
252 linkbutton component

### LINKED ISSUE #252: Create LinkButton component.
Create a component that is essentially an anchor tag that looks like a button. It should look exactly like a `Button` from `yoast-components`. See if you can create a function that applies the button styles, so you can reuse it for the `LinkButton` and no duplicate code is created.

afercia:
## Summary

* Added LinkButton component.

Noticed a pre-existing CSS issue in the buttons, where `min-height` triggers a bug in IE 11 and other CSS produced misplaced icon and text in Edge. I think we're maybe a bit too much optimistic with `flexbox`, it doesn't work so well in some browsers especially with SVG icons (not just old browsers but also Safari...)

Details on the IE bug:
https://github.com/philipwalton/flexbugs#3-min-height-on-a-flex-container-wont-apply-to-its-flex-items
min-height on a flex container won't apply to its flex items - Internet Explorer 10-11 (fixed in Edge)

IE 11:

screenshot 49

Edge:

screenshot 50

Easiest fix for the IE 11 bug: use height instead. Was there a specific reason to use min-height?

The CSS changes introduced here are not perfect, but at least the rendering is more or less consistent across browsers. Please test in depth in all browsers and platforms including IE/Edge.

Other changes:
- we don't need line-height, and it contributes to the icons misalignment (also on Safari 10)
- button elements already vertically center text natively, no need for line-height
- for links, we use `display: inline-flex;` and `align-items: center;`, treating the link content as flex items

Question:
- should we pass other link props?

## Test instructions

Use the following content in `ContentAnalysis.js` and run `yarn start`, then check localhost:3333

```
import React from "react";
import { BaseButton, Button, InlineFlexButton, IconButton } from "../../Shared/components/Button";
import { BaseLinkButton, LinkButton } from "../../Shared/components/LinkButton";

	min-height: 700px;
	padding: 40px;
	background-color: white;

	button,
	a {
		margin-right: 10px;
	}
`;

/**
 *
 */
		<h2>Buttons</h2>
		<BaseButton>Base Button</BaseButton>
		<Button>Button</Button>
		<YoastButton>Yoast Button</YoastButton>
		<YoastButton backgroundColor="lightblue" textColor="#333" withTextShadow={ false }>Yoast Button</YoastButton>
		<h2>Links</h2>
		<BaseLinkButton href="#somewhere1">Base Button</BaseLinkButton>
		<LinkButton href="#somewhere2">Button</LinkButton>
	</ContentAnalysisContainer>;
}
```

Fixes #252

### DIFF
--- a/composites/Plugin/Shared/components/Button.js
+++ b/composites/Plugin/Shared/components/Button.js
@@ -50,6 +50,7 @@ export function addBaseStyle( component ) {
 		@media all and (-ms-high-contrast: none), (-ms-high-contrast: active) {
 			// IE 10+
 			::after {
+				display: inline-block;
 				content: "";
 				min-height: ${ `${ ieMinHeight }px` };
 			}
--- a/composites/Plugin/Shared/components/Button.js
+++ b/composites/Plugin/Shared/components/Button.js
@@ -9,6 +9,30 @@ import { Icon } from "./Icon";
 import { rgba } from "../../../../style-guide/helpers";
 
 /**
+ * Returns a component with applied base button styles.
+ *
+ * @param {ReactElement} component The original component.
+ *
+ * @returns {ReactElement} Component with applied base button styles.
+ */
+export function addBaseStyle( component ) {
+	return styled( component )`
+		vertical-align: middle;
+		margin: 0;
+		padding: 4px 10px;
+		border-radius: 3px;
+		cursor: pointer;
+		box-sizing: border-box;
+		font-size: inherit;
+		font-family: inherit;
+		font-weight: inherit;
+		text-align: left;
+		height: 32px;
+		overflow: visible;
+	`;
+}
+
+/**
  * Returns a component with applied focus styles.
  *
  * @param {ReactElement} component The original component.
@@ -64,6 +88,12 @@ export function addActiveStyle( component ) {
 	`;
 }
 
+export function addFontSizeStyles( component ) {
+	return styled( component )`
+		font-size: 0.8rem;
+	`;
+}
+
 /**
  * Returns a component with all button selector styles applied.
  *
@@ -71,7 +101,7 @@ export function addActiveStyle( component ) {
  *
  * @returns {ReactElement} Component with applied styles.
  */
-export const addButtonStyles = _flow( [ addFocusStyle, addHoverStyle, addActiveStyle ] );
+export const addButtonStyles = _flow( [ addBaseStyle, addFocusStyle, addHoverStyle, addActiveStyle ] );
 
 /**
  * Returns a basic styled button.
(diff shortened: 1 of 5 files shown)


Xyfi:
@afercia The reason for the min-height was word wrapping. The whole reason we switched to flex-box is that the icon stays centred when the button contains multiple lines of text.

Regarding your question, I'm not sure what you mean since all props are passed to the anchor tag automatically.

Xyfi:
On chrome the `active` box shadow isn't connecting to the edge of the button properly. Either make the active style not show at all or add `outline: none`.

title

afercia:
Re: the focus style: this is exactly what the buttons on My Yoast look like and I've let them this way on purpose:

screen shot 2017-09-06 at 11 05 22

A focus style is always needed for accessibility reasons. When using `outline: none` (which resets the native focus style applied by browsers) then a custom focus style should be implemented. This would need a design decision. See discussion on https://github.com/Yoast/my-yoast/issues/550  where this part was discussed but not fully addressed.

afercia:
Also, the green one is the `YoastButton` ;) This PR is about the `LinkButton` (the gray ones in the second row in the screenshot above.

afercia:
Also.. this bug affects Safari 10: https://github.com/philipwalton/flexbugs#9-some-html-elements-cant-be-flex-containers
> Certain HTML elements, like <fieldset> and <button>, do not work as flex containers.

afercia:
I've spent an insane amount of time to research solutions and the only available options are:
- solution 2 mentioned in [philipwalton/flexbugs](https://github.com/philipwalton/flexbugs#3-min-height-on-a-flex-container-wont-apply-to-its-flex-items) (add a wrapper around buttons)
- use a span inside the buttons and style that as flex container
- CSS hacks to target the problematic browsers

I'd rather keep the HTML clean and don't add other elements so I've gone the CSS way.
The main issues are in Safari 10 and IE 11 and to me seemed logical to keep CSS clean for modern browsers and hack IE 11. The trick for IE 11 is to use a CSS generated pseudo element as content of the button, with a min-height equal to the button min-height subtracting borders and padding.

- all the buttons and links are now `display: inline-flex;`
- removed `InlineFlexButton` because is now useless (everything is now inline-flex)

I've tested in the following browsers, please test in all of them (and more):
Chome mac and win
Firefox mac and win
Safari 10.1.2
Safari Technology Preview
IE 11
Edge

Use the following content in ContentAnalysis.js:
```
import React from "react";
import { BaseButton, Button, InlineFlexButton, IconButton } from "../../Shared/components/Button";
import { BaseLinkButton, LinkButton } from "../../Shared/components/LinkButton";

	min-height: 700px;
	padding: 40px;
	background-color: white;

	button,
	a {
		margin-right: 10px;
	}

	.with-max-width {
		max-width: 152px;
	}
`;

/**
 *
 */
		<h2>Buttons</h2>
		<BaseButton>Base Button</BaseButton>
		<Button>Button</Button>
		<YoastButton>Yoast Button</YoastButton>
		<YoastButton backgroundColor="lightblue" textColor="#333" withTextShadow={ false }>Yoast Button</YoastButton>
		<h2>Links</h2>
		<BaseLinkButton href="#somewhere1">Base Button</BaseLinkButton>
		<LinkButton href="#somewhere2">Button</LinkButton>
	</ContentAnalysisContainer>;
}
```

Check all the buttons and links have the same height (32px) and text and icons are vertically aligned in all browsers:

screen shot 2017-09-06 at 16 04 20

Then, increase the value of `settings.minHeight` to something intentionally exaggerated to highlight potential layout issues, and check sizes and alignments again in all browsers:

screen shot 2017-09-06 at 15 30 40

afercia:
To clarify the IE 11 bug, this is how the buttons look like without the CSS hack (I've made the buttons height bigger to highlight the bug):

screenshot 65

When using `min-height` on the flex container, IE 11 is not able to make the flex items (children) aware of the container height to center them vertically.

Xyfi:
CR :+1:

herregroen:
Acceptance done 👍

PR #270, opened 2017-09-04
Added StackedProgressBar component

### LINKED ISSUE #265: Create stacked progress bar
Create a StackedPorgressBar component that can show multiple bars, each bar having a different colour and value. The component should determine the total value and the relative width of each bar.

dashboard_widget_redesign_ __7549_ _yoast_seo___waffle_io

Xyfi:
## Summary

* Adds StackProgressBar component

## Test instructions

* Replace the contents of `ContentAnalysis.js` with the following content:
```js
import React from "react";
import StackedProgressBar from "../../Shared/components/StackedProgressBar";
import colors from "../../../../style-guide/colors.json";

	box-sizing: border-box;
	min-height: 700px;
	padding: 8px;
	width: 100%;
	background-color: white;
`;

/**
 *
 */
	const items = [
		{
			value: 7,
			color: colors.$color_bad,
		},
		{
			value: 216,
			color: colors.$color_ok,
		},
		{
			value: 1134,
			color: colors.$color_good,
		},
		{
			value: 76,
			color: colors.$color_score_icon,
		},
	];
		<StackedProgressBar
			items={ items }/>
	</ContentAnalysisContainer>;
}
```

Fixes #265


composites/Plugin/Shared/components/StackedProgressBar.js
```diff
@@ -0,0 +1,59 @@
...
+};
+
+const StackedProgressBar = ( props ) => {
+	let totalValue = 0;
+	for( let i = 0; i < props.items.length; i++ ) {
+		totalValue += props.items[ i ].value;
+	}
+	return(
+		<StackedProgressBarContainer
+			aria-hidden="true" >
```
afercia: I think we don't need aria-hidden here. If I'm not wrong, these are just div/span elements with some background. Instead, in other cases where we use a `<progress>` element, we hide it because it's not well supported yet for our use case (max value different from 100).

afercia:
CR done:
the `height` prop was actually printed out in the markup producing invalid code:
`<div class="sc-bwzfXH edWTnX" height="40px">`

We should avoid to use prop names that are the same as HTML attributes. Renamed to `barHeight`.

This prop was not passed to the component so it was impossible to change the progress bar height.

`box-sizing: border-box` was a bit pointless because we're not using paddings or borders.
Fixed also a minor CSS issue: inline-block progress steps need vertical-alignment set to top.

I went ahead and applied the changes.

To test how to change the height, pass a prop with a value that includes the unit, for example:
`<StackedProgressBar barHeight="20px" ...`

or

`<StackedProgressBar barHeight="3em" ...`

I've tested also on IE 11 and Edge, looks good to me.

herregroen:
Acceptance testing done.

No major issues found. Only some slightly weird results when the total is `0` ( `width: NaN%` ) or when negative values are provided ( items can go over `100%` width ).

For completeness I'd enforce a minimum of `0` for all scores and not render any of the items if the total is `0`.

Xyfi:
CR :+1: on @herregroen 's changes.

herregroen:
Acceptance 👍  last commit was done together with @Xyfi.

PR #271, opened 2017-09-05
Stories/267 feed reader

### LINKED ISSUE #267: Implement RSS feed reader
Search for a good NPM package to read and parse RSS feeds, and write an implementation to extract the functionality needed.

It should:
* Read an RSS feed, specifically for wordpress
* Accept a number of items to retrieve
* Parse the rss feed to produce easy to use javascript objects

herregroen:
## Summary

* Adds a `getFeed` util to allow fetching and parsing RSS feeds.

* Uses https://github.com/bobby-brennan/rss-parser which seems well supported including in the browser. https://github.com/danmactough/node-feedparser is an alternative but does not seem to explicitly support the browser.

## Test instructions

* Have a Wordpress site that has CORS headers on it's RSS feed. ( You can add this to your own by adding `header("Access-Control-Allow-Origin: *");` on line 9 in `wp-includes/feed-rss2.php` ).
* Call `getFeed` with the feed URL and the maximum number of entries you wish returned. It returns a `Promise` object.
* The `Promise` should resolve with a `Feed` object as documented.

Fixes #267.


webpack.config.js
```diff
@@ -81,4 +81,7 @@ module.exports = {
 	resolve: {
 		extensions: [ ".json", ".jsx", ".js", ".svg" ],
 	},
+	node: {
```
Xyfi: If this is necessary, we should investigate another library for retrieving RSS feeds.

herregroen:
After discussion RSS feeds are now parsed using the DOMParser to significantly reduce the amount of dependencies needed.

Unfortunately this has the side-effect of making testing much harder since the DOMParser implementation for Node seems to have no support for namespaces or CDATA, both of which are used in RSS Feeds. Thus not all functionality is currently tested.

utils/getFeed.js
```diff
@@ -0,0 +1,151 @@
...
+ * @param {int}    maxItems The amount of items you wish returned, 0 for all items.
+ *
+ * @returns {Promise.<Feed>} The retrieved feed.
+ */
+export default function getFeed( url, maxItems = 0 ) {
+	return fetch( url ).then( function ( response ) {
+		return response.text();
+	} ).then( function ( raw ) {
+		return parseFeed( raw, maxItems );
+	} );
```
Xyfi: No catch clause.

herregroen: Regardless of any catch clauses a failed CORS request will always log an error in the console. This can't be avoided.

If specific error handling is required a catch method can be chained to the `getFeed` function.

I did change `parseFeed` to also return a promise with the appropriate `reject` call for errors in case a feed returns something that can't be parsed as XML.

Xyfi:
CR :+1:

boblinthorst:
Acceptance: Tested by adding the line of code to `feed-rrs2.php` as instructed, and replacing contentAnalysis.js with the following block.
```
import React from "react";
import getFeed from "../../../../utils/getFeed";

	min-height: 700px;
	width: 100%;
	background-color: white;
`;
var ding;

console.log( ding = getFeed( "http://local.wordpress.dev/feed/", 999 ) );
console.log( "console log on contentAnalysis fires." );


/**
 *
 */
	return <ContentAnalysisContainer/>;
}

```

Then, I made sure my vagrant local was running and Tested yoast-components. The feed data is retrieved and the amount specified in the documentation is shown.

PR #284, opened 2017-09-15
Add Yoast Link Button component.

### LINKED ISSUE #253: Create YoastLinkButton component
Requires #252

Create a `LinkButton` with the Yoast look and feel.

Background- and text-colour should be editable.

Don't mind the icon for now.

29705620-324c17f0-897f-11e7-8c50-687b46f81bdd_png__1632x540_

Also, implement it in the AlgolioSearcher in the `SearchResultDetail` component as the `View in KB` button. Make sure the `target="_blank"` attribute is set.

afercia:
## Summary

* Add Yoast Link Button component.

* switches everything to `display: inline-flex;`
* adds a `YoastButtonBase` component that is a button with an inner span element to fix a Safari 10 bug, see https://github.com/philipwalton/flexbugs#9-some-html-elements-cant-be-flex-containers

This turned out to be a bit more complicated than expected. Safari 10 doesn't apply correctly flexbox on `<button>` elements. The suggested fix is to add an inner `<span>` and give it the necessary flex properties. I'm styling it inheriting some CSS properties from the parent button. Worth noting this bug seems fixed in Safari Technology Preview so in the future it will be possible to remove the fix: remove `YoastButtonBase` and directly style a button element.

We might consider to apply the same workaround to the normal buttons.

To reproduce the Safari 10 bug, use the Safari 10 inspector and remove the span from the "Test Button".

Not particularly happy with this, please let me know if the workaround is acceptable, any feedback welcome.

## Test instructions

* Put this in ContentAnalysis.js, run `yarn start` and go to http://localhost:3333
```
import React from "react";
import { BaseButton, Button, IconButton } from "../../Shared/components/Button";
import { BaseLinkButton, LinkButton } from "../../Shared/components/LinkButton";
import { YoastLinkButton } from "../../Shared/components/YoastLinkButton";

	min-height: 700px;
	padding: 40px;
	background-color: white;

	button,
	a {
		margin-right: 10px;
	}

	.with-max-width {
		max-width: 152px;
	}

	.test-large-button {
		min-width: 200px;
	}
`;

/**
 *
 */
		<h2>Buttons</h2>
		<BaseButton>Base Button</BaseButton>
		<Button>Button</Button>
		<YoastButton onClick={ () => {
			console.log( "hello Button" );
		} }>Yoast Button</YoastButton>
		<YoastButton backgroundColor="lightblue" textColor="#333" withTextShadow={ false }>Yoast Button</YoastButton>
		<YoastButton className="test-large-button">Test Button</YoastButton>
		<h2>Links</h2>
		<BaseLinkButton href="#somewhere1">Base Button</BaseLinkButton>
		<LinkButton href="#somewhere2">Button</LinkButton>
		<YoastLinkButton href="#somewhere3">Yoast Button</YoastLinkButton>
		<YoastLinkButton href="#somewhere4" backgroundColor="lightblue" textColor="#333" withTextShadow={ false }>Yoast Button</YoastLinkButton>
		<YoastLinkButton className="test-large-button" href="#somewhere3">Test Button</YoastLinkButton>
	</ContentAnalysisContainer>;
}
```

Fixes #253

### DIFF
--- a/composites/Plugin/Shared/components/YoastButton.js
+++ b/composites/Plugin/Shared/components/YoastButton.js
@@ -1,8 +1,17 @@
 import React from "react";
 import PropTypes from "prop-types";
 import styled from "styled-components";
+
 import colors from "../../../../style-guide/colors.json";
 import { rgba } from "../../../../style-guide/helpers";
+
+const settings = {
+	minHeight: 48,
+	verticalPadding: 0,
+	borderWidth: 0,
+};
+
+const ieMinHeight = settings.minHeight - ( settings.verticalPadding * 2 ) - ( settings.borderWidth * 2 );
 
 /**
  * Returns a component with a Yoast button-like style.
@@ -20,9 +29,10 @@ export function addButtonStyles( component ) {
 		align-items: center;
 		justify-content: center;
 		vertical-align: middle;
-		height: 48px;
+		min-height: ${ `${ settings.minHeight }px` };
 		margin: 0;
 		padding: 0 16px;
+		padding: ${ `${ settings.verticalPadding }px` } 16px;
 		border: 0;
 		border-radius: 4px;
 		box-sizing: border-box;
@@ -49,6 +59,15 @@ export function addButtonStyles( component ) {
 			align-items: inherit;
 			justify-content: inherit;
 			width: 100%;
+		}
+
+		// Only needed for IE 10+.
+		@media all and (-ms-high-contrast: none), (-ms-high-contrast: active) {
+			::after {
+				display: inline-block;
+				content: "";
+				min-height: ${ `${ ieMinHeight }px` };
+			}
 		}
 	`;
 }
--- a/composites/Plugin/Shared/components/YoastButton.js
+++ b/composites/Plugin/Shared/components/YoastButton.js
@@ -1,3 +1,4 @@
+import React from "react";
 import PropTypes from "prop-types";
 import styled from "styled-components";
 import colors from "../../../../style-guide/colors.json";
@@ -15,12 +16,14 @@ import { rgba } from "../../../../style-guide/helpers";
  */
 export function addButtonStyles( component ) {
 	return styled( component )`
-		display: inline-block;
+		display: inline-flex;
+		align-items: center;
+		justify-content: center;
+		vertical-align: middle;
 		height: 48px;
 		margin: 0;
 		padding: 0 16px;
 		border: 0;
-		vertical-align: middle;
 		border-radius: 4px;
 		box-sizing: border-box;
 		font: 400 14px/24px "Open Sans", sans-serif;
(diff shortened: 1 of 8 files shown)


afercia:
About:
>  implement it in the AlgoliaSearcher in the SearchResultDetail component as the View in KB button. Make sure the target="_blank" attribute is set.

of course, it needs #243

composites/Plugin/Shared/components/YoastButton.js
```diff
@@ -39,18 +42,58 @@ export function addButtonStyles( component ) {
...
+ * Returns a component with a button and an inner span element.
+ *
+ * The inner span is only needed to fix a Safari 10 bug with flexbox and button
+ * elements. This bug is fixed in Safari Technology Previs and in the future it
+ * will be possible to remove this component and directly style a button element.
+ * See https://github.com/philipwalton/flexbugs#9-some-html-elements-cant-be-flex-containers
+ *
+ * @returns {ReactElement} The button with inner span.
+ */
+const YoastButtonBase = ( { className, onClick, type, children } ) => (
```
Xyfi: Now you are limiting the amount of attributes that can be passed to the button. Isn't it better to deconstruct the props into the button? (`<button { ...props } >`)

composites/Plugin/Shared/components/YoastButton.js
```diff
@@ -39,18 +42,58 @@ export function addButtonStyles( component ) {
...
+};
+
+YoastButtonBase.defaultProps = {
+	type: "button",
+};
+
+/**
  * Returns a Button with the Yoast button style.
  *
+ * We're styling the YoastButtonBase component just because of the Safari 10 bug.
```
Xyfi: Maybe create a reference to the bug you mean (e.g. `(see YoastButtonBase JSDoc)`)

afercia:
For documentation:
the min-height bug in IE 11 we faced for the normal buttons is a different story. As soon as a button has a min-height, IE triggers the bug and the `<span>` introduced here doesn't help to fix it. 

This means that, if we want buttons with a min-height, the only workaround so far is the trick with `::after` (the red vertical thing visible in the normal buttons in the screenshot below):

screenshot 79

composites/Plugin/Shared/components/YoastLinkButton.js
```diff
@@ -0,0 +1,33 @@
+import PropTypes from "prop-types";
+import styled from "styled-components";
+import colors from "../../../../style-guide/colors.json";
+import { addButtonStyles } from "./YoastButton";
+
+/**
+ * Returns a Button with the Yoast button style.
```
Xyfi: Technically not a button.

composites/Plugin/Shared/components/YoastLinkButton.js
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * @param {object} props Component props.
+ *
+ * @returns {ReactElement} Styled button.
```
Xyfi: Technically not a button.

composites/Plugin/Shared/components/YoastButton.js
```diff
@@ -15,12 +16,14 @@ import { rgba } from "../../../../style-guide/helpers";
  */
 export function addButtonStyles( component ) {
 	return styled( component )`
```
Xyfi: Does this mean this function now applies styles that are specific to `button` and `a` elements?

composites/Plugin/Shared/components/YoastButton.js
afercia: That will throw warnings like:
`Unknown props `backgroundColor`, `textColor`, `withTextShadow` on <button> tag.`

composites/Plugin/Shared/components/YoastButton.js
afercia: Yes and it's basically the same thing we're already doing for `BaseLinkButton`.

afercia:
Addressed feedback about docs, see comments for the other points.

afercia:
Latest commits add the same IE 10+ CSS hack already used for the normal buttons and add a test.

This should be tested in IE 11, setting a higher value for the buttons min-height. In the screenshot below, I've made the CSS pseudo element visible to give an idea about how the CSS hack works:

screen shot 2017-09-18 at 13 45 19

Xyfi:
CR :+1: Tested on IE, Edge, Chrome, Firefox, Safari, Chrome on Android.

IreneStr:
Acceptance 👍  Tested on iOs for iPhone, Opera and Chrome.

PR #290, opened 2017-09-19
Stories/249 wizard css tweaks

### LINKED ISSUE #249: Onboarding Wizard Various CSS tweaks
From https://github.com/Yoast/wordpress-seo/issues/7744:
> In `yoast-components/composites/OnboardingWizard/onboarding-wizard.scss`:
> 
> - [ ] Change `&--step&__active svg` to `$color_green`.
> - [ ] Change `.#{$prefix}` padding from 3em to 2em.
> - [ ] Change/add these properties to `&--step--container h2`:
>    - [ ] change font size to 2.25em.
>    - [ ] change font weight to 100.
>    - [ ] add `margin: 0`.
> 
> In `wordpress-seo/css/src/yoast-components.scss`:
> - [ ] Change `.yoast-wizard &-body` max-width from 80em to 60em.
> 
> In _somewhere I couldn't find_:
> - [ ] Change the heading of each step from an `<h1>` to an `<h2>`
> - [ ] Change "Stay up-to-date" in the newsletter step from an `<h2>` to an `<h3>`.
> 
> **As a side note**, it's weird that there are CSS properties related to the wizard scattered throughout different files and even different repos, as the above shows. This should be unified if possible.

As the last line suggests, try to define the styling in `yoast-components` as much as possible.

Xyfi:
## Summary

* Onboarding Wizard styling changes

## Test instructions

* Open the onboarding wizard
* Check if the CSS changes described in the issue have been applied

Fixes #249


afercia:
> Change the heading of each step from an <h1> to an <h2>

I think this was requested because the new logo in the wizard was initially supposed to be a h1, so moving all the other headings down to h2 made perfectly sense. We then changed our mind and made the new logo a simple image with a paragraph (no h1) so I think the headings in the steps should stay as h1. /cc @hedgefield

composites/OnboardingWizard/onboarding-wizard.scss
```diff
@@ -121,8 +121,8 @@ body {
 			}
 
 			svg {
-				color: $color_purple;
-				fill: $color_purple;
+				color: $color_green;
+				fill: $color_green;
```
afercia: Seem these properties don't do anything, a bit pointless even before the current changes. As far as I see, the step numbers color is passed in the configuration and applied as inline styles so the green won't be applied:

screen shot 2017-09-20 at 09 53 08

Re: `$color_green`, it has a very low contrast ratio with white: https://jdlsn.com/color/?type=hex&color=FFFFFF&color2=77b227. The result is 2.57:1 where it should be at least 4.5:1. Our greens are a bit tricky for a11y and we faced this issue also in other places. Maybe worth considering a global review of the greens. In this specific case, I'd suggest to use $color_green_medium     #008a00

afercia:
ABout the step buttons color, currently they are based on a "material UI theme" config, see `composites/OnboardingWizard/config/yoast-theme.js`

where the `primary1Color` color is used also for the next button. See also http://www.material-ui.com/#/customization/themes

afercia:
CR done 🚧 

Thinking the headings should stay unchanged. The step buttons color should be passed to the material ui config, as `stepper.iconColor`.

afercia:
In the last commit:
- pass the active step button color to the material UI theme config
- use a darker green
- restore the previous heading levels
- add max-width to the `#{ $prefix }-body` (should be removed from wordpress-seo)

afercia:
Re: `$color_green`, it has a very low contrast ratio with white: https://jdlsn.com/color/?type=hex&color=FFFFFF&color2=77b227. The result is 2.57:1 where it should be at least 4.5:1. Our greens are a bit tricky for a11y and we faced this issue in other places too. Maybe worth considering a global review of the greens. In this specific case, I've changed it to $color_green_medium     #008a00

PR #293, opened 2017-09-20
Fix justify-content & z-index prop warning on LoadingIndicator component

### LINKED ISSUE #292: LoadingIndicator: unknown props justify-content, z-index on div tag
To reproduce:
- start the yoast components
- go to the Wizard
- click on the second step (so the loading indicator is rendered)
- observe your browser's console:

`Warning: Unknown props 'justify-content', 'z-index' on <div> tag. Remove these props from the element.`

See:
```
const LoadingIndicator = () => (
	<div className="yoast-wizard-overlay-loader">
		<RefreshIndicator
			size={40}
			left={10}
			top={100}
			status="loading"
			style={style.refresh}
			justify-content= "center"
			z-index={11}
		/>
	</div>
);
```

these properties are passed directly to a div element, maybe they should be passed to the style object instead. Even with that, I'm not sure what `justify-content` is supposed to do since, as far as I see, the loading indicator is not a flex container.

Xyfi:
## Summary

* Fixes a console error when switching pages in the Onboarding wizard.

## Test instructions

### To check that the prop warnings are gone
1. Open the OnboardingWizard and go to the next step.
2. Verify that the prop warning is no longer showing up in the console.

### To check that the props are properly applied as styles
1. Place `debugger;` as the first line in the `handleSuccessful` function in `OnboardingWizard.js`.
2. Open the OnboardingWizard and make sure you have the chrome debug tools open.
3. Go to the next step. Wait for the code execution to be paused by the `debugger` statement.
4. Search for the div where the loading indicator is rendered. It's **inside** a div with the class `"yoast-wizard-overlay-loader"`. Verify that the styles have been applied to this div.

```css
justify-content: "center";
z-index: 11;
```

Fixes #292

### DIFF
--- a/composites/OnboardingWizard/LoadingIndicator.js
+++ b/composites/OnboardingWizard/LoadingIndicator.js
@@ -10,8 +10,6 @@ const style = {
 	refresh: {
 		display: "inline-block",
 		position: "relative",
-		zIndex: 11,
-		justifyContent: "center",
 	},
 };
 
--- a/composites/OnboardingWizard/LoadingIndicator.js
+++ b/composites/OnboardingWizard/LoadingIndicator.js
@@ -10,6 +10,8 @@ const style = {
 	refresh: {
 		display: "inline-block",
 		position: "relative",
+		zIndex: 11,
+		justifyContent: "center",
 	},
 };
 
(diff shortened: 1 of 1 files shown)


afercia:
Actually, the styles were correctly removed, since they don't do anything.

afercia:
CR and acceptance 👍 

Noticed a layout bug in IE 11 which is unrelated to this change, will open a new issue.

PR #302, opened 2017-09-25
A11yNotice

### LINKED ISSUE #278: Add A11yNotice component
Create a component that will read messages to screen readers but will not display to other users.

Requirements: 
* Should be able to translate messages using `react-intl`.

The css for this component can be found here:
https://github.com/wpaccessibility/a11ythemepatterns/blob/master/read-more-links/style.css

Be sure to replace the current solution use in `WordpressFeed.js`, after implementation.

Xyfi:
## Summary

* Adds an invisible A11y component for screen readers.

* Made the component a `span` element by default, but this can be overwritten with `styled-components`'s [`.withComponent`](https://www.styled-components.com/docs/api#withcomponent) function.

## Test instructions

* Copy the following content into `ContentAnalysis.js`:
```js
import React from "react";

import { A11yNotice } from "../../Shared/components/A11yNotice";

	min-height: 700px;
	width: 100%;
	background-color: white;
`;

/**
 *
 */
	const H1A11yNotice = A11yNotice.withComponent( "h1" );
		<p>The following text is invisible:</p>
		<A11yNotice>Invisible text</A11yNotice>
		<p>The following header is invisible:</p>
		<H1A11yNotice>Invisible header</H1A11yNotice>
	</ContentAnalysisContainer>;
}
```
* Inspect the two `p` tags that are displayed and make sure they are followed by a invisible `span` and `h1` element.
* Use a screen reader to make sure the text is read to you.

Fixes #278

### DIFF
--- a/composites/HelpCenter/views/VideoTutorial.js
+++ b/composites/HelpCenter/views/VideoTutorial.js
@@ -1,0 +1,48 @@
+import React from "react";
+import styled from "styled-components";
+import YouTubeVideo from "../../basic/YouTubeVideo";
+import PropTypes from "prop-types";
+
+const Section = styled.section`
+	display: flex;
+	padding: 1em;
+`;
+
+const VideoTutorialPlaceholder = styled( Section )`
+	flex-wrap: wrap;
+`;
+
+const VideoTextPanel = styled( Section )`
+	min-width: 220px;
+	max-width: 550px;
+	flex: 50% 0;
+`;
+
+const StyledYouTubeVideo = styled( YouTubeVideo )`
+	flex: 50% 0;
+`;
+
+/**
+ * Creates a VideoTutorial component that displays a YouTube video.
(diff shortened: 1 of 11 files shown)


composites/HelpCenter/views/VideoTutorial.js
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * @param {Object} props The props to use for the component.
+ * @returns {ReactElement} The rendered VideoTutorial component.
+ *
+ * @constructor
+ */
+export default function VideoTutorial( props ) {
+	return (
+		<VideoTutorialPlaceholder>
+			<StyledYouTubeVideo	src={ props.src } title={ props.title } />
```
boblinthorst: tab indent, should be space.

Xyfi:
New PR for this issue #306

PR #294, opened 2017-09-20
Add Tabs and SubNavigation components

IreneStr:
## Summary

This PR can be summarized in the following changelog entry:

* Adds Tabs and SubNavigation components

## Relevant technical choices:

* I've moved these files to the basic folder and made the name more generic than in @jcomack's original PR.

## Test instructions

Copy this to ContentAnalysis.js:

````js
import React from "react";
import styled from "styled-components";

import SubNavigation from "../../../basic/SubNavigation";
import PreviousTabs from "../../../basic/Tabs";
import YoastTabs from "../../../basic/YoastTabs";

export const ContentAnalysisContainer = styled.div`
	min-height: 700px;
	padding: 40px;
	background-color: white;
`;

let items = [
	{
		id: "tab1",
		label: "tab1",
		url: "http://localhost:3333/tab1",
		view: <p>This is some content for tab 1. <a href="#">focusable element 1</a></p>,
	},
	{
		id: "tab2",
		label: "tab2",
		url: "http://localhost:3333/tab2",
		view: <p>This is some content for tab 2. <a href="#">focusable element 2</a></p>,
	},
	{
		id: "tab3",
		label: "tab3",
		url: "http://localhost:3333/tab3",
		view: <p>This is some content for tab 3. <a href="#">focusable element 3</a></p>,
	},
];

/**
 * Returns the ContentAnalysis component.
 *
 * @returns {ReactElement} The ContentAnalysis component.
 */
export default function ContentAnalysis() {
	return <ContentAnalysisContainer>
		<h1>Tabs experiments</h1>
		<h2>Previous implementation</h2>
		<SubNavigation items={ items } activeTab="http://localhost:3333/tab1" onClick={ () => {} } />
		<PreviousTabs items={ items } activeTab="http://localhost:3333/tab1" />
		<h2>Using "react-tabs"</h2>
		<YoastTabs items={ items } />
	</ContentAnalysisContainer>;
}
````

Fixes #215
Fixes #216

PR #303, opened 2017-09-25
Introduce react-tabs for the Help Center tabs.

### LINKED ISSUE #216: Implement HelpCenterTabs component
- Implement HelpCenterTabs component

Part 6 of the implementation of pull #203

afercia:
## Summary

* Adds the Help Center ARIA tabs widget component

* Introduces `react-tabs`

## Test instructions

* use the following code in ContentAnalysis
```
import React from "react";

import YoastTabs from "../../Shared/components/YoastTabs";

	min-height: 700px;
	padding: 40px;
	background-color: white;
`;

let items = [
	{
		id: "tab1",
		label: "video tutorial",
		content: <p>This is some content for tab 1. <a href="#">focusable element 1</a></p>,
	},
	{
		id: "tab2",
		label: "knowledge base",
		content: <p>This is some content for tab 2. <a href="#">focusable element 2</a></p>,
	},
	{
		id: "tab3",
		label: "email support",
		content: <p>This is some content for tab 3. <a href="#">focusable element 3</a></p>,
	},
];

/**
 *
 */
		<h2>Using "react-tabs"</h2>
		<YoastTabs items={ items } />
	</ContentAnalysisContainer>;
}
```

Questions:
- does `view` need to be named `view`? for consistency with the ScoreAssessments, `html` would be better or, maybe, `content` even better
- does it needs to be passed as html object (as implemented in the original PR see #294) or as string? In the second case, I guess it should use `dangerouslySetInnerHTML` ...

Fixes #216

### DIFF
--- a/composites/Plugin/Shared/tests/__snapshots__/YoastTabsTest.js.snap
+++ b/composites/Plugin/Shared/tests/__snapshots__/YoastTabsTest.js.snap
@@ -1,0 +1,86 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`the YoastTabs matches the snapshot 1`] = `
+<div
+  className="sc-bdVaJa kXyDjN"
+>
+  <div
+    className="react-tabs"
+    data-tabs={true}
+    onClick={[Function]}
+    onKeyDown={[Function]}
+  >
+    <ul
+      className="react-tabs__tab-list"
+      role="tablist"
+    >
+      <li
+        aria-controls="react-tabs-1"
+        aria-disabled="false"
+        aria-selected="true"
+        className="react-tabs__tab react-tabs__tab--selected"
+        id="react-tabs-0"
+        role="tab"
+        tabIndex="0"
+      >
+        tab1
+      </li>
+      <li
+        aria-controls="react-tabs-3"
+        aria-disabled="false"
+        aria-selected="false"
+        className="react-tabs__tab"
+        id="react-tabs-2"
+        role="tab"
+        tabIndex={null}
+      >
+        tab2
+      </li>
+      <li
+        aria-controls="react-tabs-5"
+        aria-disabled="false"
+        aria-selected="false"
+        className="react-tabs__tab"
+        id="react-tabs-4"
+        role="tab"
+        tabIndex={null}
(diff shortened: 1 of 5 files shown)


afercia:
From an a11y perspective, tabs should work this way:
- only one tab is focusable at a time
- once a tab is focused, navigation through tabs should work with arrow keys only
- instead, tabbing from a focused tab should bring directly to the associated tab panel

screen readers should announce the number of tabs and the currently selected one:

screen shot 2017-09-25 at 14 47 21

when tabbing from a tab to its tabpanel, the tabpanel should get focus and screen readers should announce its name, which is given by `aria-labelledby` that targets the associated tab above the tabpanel:

screen shot 2017-09-25 at 14 47 27

This mix of navigation based on the Tab key and the arrow keys is intentional and meant to provide fast navigation through complex UIs. It is often used also in other ARIA widgets.

ARIA widgets are often based on already existing interfaces in our operating systems, and they try to replicate their interaction. See https://github.com/Yoast/wordpress-seo/issues/5172 with some screenshots of the macOS system preferences UI, which uses tabs. On Windows, a typical example is the Task manager.

afercia:
Renamed `view` to `content` as agreed with @IreneStr and @jcomack

Xyfi:
CR :+1:

I made some minor changes.

composites/Plugin/Shared/tests/__snapshots__/YoastTabsTest.js.snap
```diff
@@ -3,84 +3,80 @@
 exports[`the YoastTabs matches the snapshot 1`] = `
 <div
   className="sc-bdVaJa kXyDjN"
+  data-tabs={true}
+  onClick={[Function]}
+  onKeyDown={[Function]}
 >
-  <div
-    className="react-tabs"
```
IreneStr: With this commit, the "react-tabs" className has been lost. Is that a problem?

IreneStr: Discussed with @Xyfi. Seems not to be a problem.

IreneStr:
Acceptance 🆗

PR #310, opened 2017-09-27
Help center

### LINKED ISSUE #217: Implement HelpCenter component
- Implement HelpCenter component
- Add HelpCenter to index.js
- Add HelpCenter to App.js
- Rename `Title` to `Yoast Components` in index.html

Part 7 of the implementation of pull #203 

When this issue is closed, please also close the original, big PR #203

Xyfi:
## Summary

* Added Help Center component.

## Test instructions

* In app.js, pass the following items to the HelpCenterWrapper:
````js
let items = [
	{
		label: "Video tutorial",
		id: "video-tutorial",
		content: <div>Video tutorial content.</div>,
	},
	{

		label: "Knowledge base",
		id: "knowledge-base",
		content: <div>KB content.</div>,
	},
	{
		label: "Get support",
		id: "support",
		content: <div>Get support content.</div>,
	},
];
````

[Edited by Andrea]

## Specs to meet

> The default (min-)height of the paper is 432px (give or take, I know this depends on the font size and paddings/margins eventually). Including the toggle above it it is ~500px.
> 
> The font size should be inherited from the default body text.
> 
> The padding is 40px left and right, and 24px top and bottom.

Fixes #217

### CODE REVIEW

**Comment by afercia** [comment]
Current tabs alignment:

screen shot 2017-10-02 at 09 24 31

Notice the overall effect looks weird because the central tab text is not in the center of the page. The tabs have a different width, which is given by their content. And this is how flexbox works: it distributes content evenly (left and right margins are the same). To match the design (see below):

help kb

we need to give the tabs the same width which in flexbox terms is the flex-basis value. Alternatively, a same min-width.

Not sure the tabs text should be purple, in the design is grey.

Not sure the 432px min-height should include the toggle button:
> the default (min-)height of the paper is 432px

to my understanding, the "paper" is the white container around the tabs.

I'd like to see #324 merged before addressing these issues.

**Comment by afercia** [comment]
CR 🚧 
the design need to match the mockup a bit better
I guess we should also add a "paper" around the help center
the button text "Need help?" needs to be translatable

**Comment by afercia** [comment]
Current rendering (remove the button props to see the button in the green version):

screen shot 2017-10-02 at 13 23 26

Todo: need to discuss how to have the tabs text/color style be different in My Yoast and in the plugin.

**Comment by afercia** [comment]
The issue with this component is that the provided design requires different... design on the Plugin and on My Yoast:

31076430-5ef5e3f8-a77a-11e7-86fa-6ef7222ec505

28922610-2eebbfbe-785b-11e7-8686-2a34ba6bc5f9

While the functionality is the same, the toggle button and tabs styles are completely different. I'd note this defeats the purpose of having re-usable components, since to match the 2 designs there's the need to pass a big number of props to control the styles (or build completely separate components, but in any case their styles should be controlled).

There are now the following properties on the Help Center, and some of them get passed to the YoastTabs:
```
buttonBackgroundColor={ colors.$color_white }
buttonTextColor={ colors.$color_pink_dark }
buttonIconColor={ colors.$color_pink_dark }
buttonWithTextShadow={ false }
tabsTextColor={ colors.$color_pink_dark }
tabsFontSize="1em"
tabsTextTransform="uppercase"
tabsFontWeight="normal"
tabsBaseWidth="200px"
```

Remove all of them from `app/HelpCenterWrapper.js` to see the My Yoast style.

Two important things to define with design /cc @hedgefield 

The hover/focus on the toggle button currently inherits from the YoastButton, not so nice when the button is white

The vertical centering of text inside the tabs: this is particularly tricky since it depends on the actual font that will be used. On My Yoast the font is Open Sans, on the plugin the font will be the system font of the used operating system.
Also, on the Plugin, the uppercase text is notoriously difficult to align and would require fine-tuning the top and bottom paddings... that would mean two more props to pass... let me know if having a perfect vertical alignment is important

**Comment by hedgefield** [comment]
@afercia thanks for nothing this. The two mockups are from different times, so the design may differ, but here in the real world, you may try to standardize as much of the things as possible :) The tabs should have no difference aside from the font used (I don't care much about the alignment). The toggle can be green as on my yoast, I discussed that earlier today with @IreneStr. And notably the search bar designs in the above mockups differ a bit (icon missing). Those should of course also be the same. Hope that helps!

**Comment by afercia** [comment]
@hedgefield 
> The tabs should have no difference aside from the font used

the tabs text have at least these differences:
- color
- font size
- uppercase/normal
- font weight 

please specify which ones of these can be standardized.

**Comment by hedgefield** [comment]
Take them all from the my yoast version: grey, 24px, regular weight, lowercase (or, well, _Capitalized_).

**Comment by afercia** [comment]
Great. That's the default style. To test: just don't pass the props related to the tabs:
```
				tabsTextColor={ colors.$color_pink_dark }
				tabsFontSize="1em"
				tabsTextTransform="uppercase"
				tabsFontWeight="normal"
				tabsBaseWidth="200px"
```

**Comment by afercia** [comment]
Sorry just to clarify `regular weight` seems to me it's actually a light weight.

**Comment by hedgefield** [comment]
Ah good call @afercia, it is, go with that ;)

**Comment by afercia** [comment]
Checking in IE 11, the button has the vertical alignment bug. Looking at why the implemented fix doesn't work.

screenshot 87

**Comment by afercia** [comment]
Weird but the CSS media query hack for IE 11 works only if there are no spaces within parenthesis:
`@media all and (-ms-high-contrast: none), (-ms-high-contrast: active)`

they were recently added as coding standards in https://github.com/Yoast/yoast-components/commit/76a3650f9f792e2057b499d769110bf3365e0532

**Comment by afercia** [comment]
Tested on all Windows browsers, look good to me. The only thing left, maybe, is the hover/focus style which draws a inner box shadow making the button darker. This looks good on the green button, maybe not so much on the white button. /cc @hedgefield 

screen shot 2017-10-02 at 17 17 23

**Comment by hedgefield** [comment]
Ah well, that's a minor detail that I don't immediately see a better solution for, wouldn't spend any time worrying about it 👍

File: app/HelpCenterWrapper.js [hunk 1]
```diff
@@ -0,0 +1,69 @@
+import React from "react";
+import styled from "styled-components";
+
+import HelpCenter from "../composites/Plugin/HelpCenter/HelpCenter";
+import VideoTutorial from "../composites/HelpCenter/views/VideoTutorial";
+import AlgoliaSearcher from "../composites/AlgoliaSearch/AlgoliaSearcher";
+import colors from "../style-guide/colors.json";
+
+export const HelpCenterContainer = styled.div`
+	background-color: #f1f1f1;
```
IreneStr: $color_grey_light

app/HelpCenterWrapper.js
```diff
@@ -0,0 +1,69 @@
...
+`;
+
+/**
+ * Returns the HelpCenterWrapper component.
+ *
+ * @param {Object} props The props to use for the component.
+ *
+ * @returns {ReactElement} The HelpCenterWrapper component.
+ */
+export default function HelpCenterWrapper( props ) {
```
IreneStr: Props are not used

boblinthorst:
Code review on last three commits by Irene, done 👍

IreneStr:
Acceptance :+1:

Issue #196, opened 2017-08-03
Reactify Help Center - invocation and navigation


hedgefield:
The way the new Help Center will be invoked is very similar to the current design.

- There is a button that toggles the display of the paper that holds the Help Center.
- Within the paper is a horizontal tab navigation that toggles the display of the individual pages.

![help closed](https://user-images.githubusercontent.com/6073772/28922605-2ada3626-785b-11e7-8f5e-8f40580e2c23.png)

![help kb](https://user-images.githubusercontent.com/6073772/28922610-2eebbfbe-785b-11e7-8686-2a34ba6bc5f9.png)

## The button
The button is of the same class as all the other My Yoast buttons. It has an icon (fa-question-circle), a label (`Need help?`) and a chevron that points down when the HC is closed, and points up when the HC is open.

The button is horizontally aligned to the middle of the content column. In My Yoast, it is vertically aligned to the top of the page. In the plugin, it appears in the same place it does now (but also middle-aligned).

If possible, clicking the button should animate the opening of the help center paper, it shouldn't just appear. It's fine if the paper is blank during this transition.

## The tabs
The tab navigation is the same class as the other tabs in My Yoast. It holds three items: `Video tutorial`, `Knowledge base` and `Get support`. Note: on the Titles & Meta page in the plugin, it has a fourth tab: `Template explanation`, the contents of which are described in a separate issue. Therefore the tabs should not have a fixed width (if they didn't already).

## The paper
The help center container itself should stretch to 100% of the content column, using the same margin that the rest of the content uses. I hope this is clear, but when in doubt, come and ask me.

The default (min-)height of the paper is 432px (give or take, I know this depends on the font size and paddings/margins eventually). Including the toggle above it it is ~500px.

The font size should be inherited from the default body text.

The padding is 40px left and right, and 24px top and bottom.

afercia:
Re: Tabs:
this would be a good opportunity to build an ARIA tabbed interface component. Interaction details and code examples (sorry, no React): https://www.w3.org/TR/wai-aria-practices/#tabpanel

afercia:
Of course, _Google knows_ so [from a very quick search](https://www.google.it/search?q=react+aria+tabs&oq=react+aria+tabs&aqs=chrome..69i57.2551j0j7&sourceid=chrome&ie=UTF-8) seems there are already some React accessible tabs components:
https://github.com/reactjs/react-tabs
https://github.com/stowball/react-accessible-tabs

no demo, need testing
[...]

PR #318, opened 2017-09-28
Add Help Center toggle button.

### LINKED ISSUE #311: Implement Help Center Toggle Button
Implement the Help Center Toggle Button Component.

help closed

## The button
The button is of the same class as all the other My Yoast buttons. It has an icon (fa-question-circle), a label (`Need help?`) and a chevron that points down when the HC is closed, and points up when the HC is open.

The button is horizontally aligned to the middle of the content column. In My Yoast, it is vertically aligned to the top of the page. In the plugin, it appears in the same place it does now (but also middle-aligned).

If possible, clicking the button should animate the opening of the help center paper, it shouldn't just appear. It's fine if the paper is blank during this transition.

(Above description copied from #196)

afercia:
## Summary

* Add Help Center toggle button component

* this button seems to me a unique case, so I've put it in its own file
* can receive an `isExpanded` prop from a wrapper component
* added `isExpanded` prop to the base YoastButton
* changed the `children` prop to accept, array of nodes, node, string

**Question:**
worth noting in the design in https://github.com/Yoast/yoast-components/issues/311 and https://github.com/Yoast/yoast-components/issues/196 this button seems to have squared corners even if the comment says "The button is of the same class as all the other My Yoast buttons". However, the other buttons have rounded corners... not a big deal, `border-radius` can be reset but needs clarification from design. /cc @hedgefield 

## Test instructions

* test with the following code in ContentAnalysis and check the new button matches the design
```
iimport React from "react";

import { IconButton } from "../../Shared/components/Button";
import { questionCircle } from "../../../../style-guide/svg";
import { YoastLinkButton } from "../../Shared/components/YoastLinkButton";
import { HelpCenterButton } from "../../Shared/components/HelpCenterButton";
import { makeOutboundLink } from "../../../../utils/makeOutboundLink";

	min-height: 700px;
	width: 100%;
	background-color: white;

	button {
		margin-right: 16px;
	}
`;

const RightYoastLinkButton = makeOutboundLink( styled( YoastLinkButton )`
	float: right;
` );

/**
 *
 */
		<IconButton icon={ questionCircle } iconColor="#f00">Icon Button</IconButton>
		<YoastButton>Button</YoastButton>
		<HelpCenterButton onClick={ () => {
			console.log( "Toggle me!" );
		} }>Need Help?</HelpCenterButton>
		<RightYoastLinkButton href="#somewhere">Children string</RightYoastLinkButton>
		<RightYoastLinkButton href="#somewhere"><span>Children node</span></RightYoastLinkButton>
	</ContentAnalysisContainer>;
}
```

current rendering:

screen shot 2017-09-28 at 16 11 15

Fixes #311


hedgefield:
Nice. The square corners were a holdover from previous designs where it was more of a feedback label, and because it lined up with the edge of the screen, but it looks good like this and makes much more sense because _it is_ a button, so keep the radius :)

composites/AlgoliaSearch/tests/__snapshots__/SearchResultDetailTest.js.snap
```diff
@@ -8,6 +8,7 @@ exports[`the SearchResultDetail component matches the snapshot 1`] = `
     className="sc-gzVnrw kJqUwD"
   >
     <button
+      aria-expanded={undefined}
```
IreneStr: undefined

composites/Plugin/Shared/components/HelpCenterButton.js
```diff
@@ -0,0 +1,49 @@
...
+			{ props.children }
+			<Icon icon={ props.isExpanded ? angleUp : angleDown } color={ props.iconColor } />
+		</YoastButton>
+	);
+};
+
+export const HelpCenterButton = styled( HelpCenterButtonBase )`
+	min-width: 0;
+	padding: 0 2px;
+
```
IreneStr: newline needed?

composites/Plugin/Shared/components/YoastButton.js
```diff
@@ -86,7 +86,12 @@ YoastButtonBase.propTypes = {
 	className: PropTypes.string,
 	onClick: PropTypes.func,
 	type: PropTypes.string,
-	children: PropTypes.node,
+	isExpanded: PropTypes.bool,
+	children: PropTypes.oneOfType( [
+		PropTypes.arrayOf( PropTypes.node ),
+		PropTypes.node,
+		PropTypes.string,
```
IreneStr: Aren't strings nodes?

composites/AlgoliaSearch/tests/__snapshots__/SearchResultDetailTest.js.snap
afercia: I think this is normal in the snapshots, and just means the prop hasn't been passed. Better ask @Xyfi  though. (there are other similar occurrences in the snapshots)

screen shot 2017-09-29 at 12 53 22

composites/Plugin/Shared/components/HelpCenterButton.js
afercia: In plain CSS, `svg { ...` would be a new selector and yes, it would need a new line to be separated from the previous rule. In styled components, it's just a matter of readability I guess, and internal conventions.

composites/Plugin/Shared/components/YoastButton.js
afercia: Not sure in React terms. In JavaScript terms, no they're not the same thing. In DOM terms, `text` is a node. See also `makeOutboundLink`:
```
		children: PropTypes.oneOfType( [
			PropTypes.string,
			PropTypes.node,
		] ),
```

composites/AlgoliaSearch/tests/__snapshots__/SearchResultDetailTest.js.snap
IreneStr: Let me be more specific 😄 : Why isn't the prop passed in the test?
To me it seems to be good practice to not have `undefined` values in the tests, because you wouldn't want `undefined`s in the eventual implementation.

composites/Plugin/Shared/components/YoastButton.js
IreneStr: According to Facebook's React docs:
```
// Anything that can be rendered: numbers, strings, elements or an array
// (or fragment) containing these types.
  optionalNode: PropTypes.node,
```

composites/Plugin/Shared/tests/__snapshots__/HelpCenterButtonTest.js.snap
```diff
@@ -0,0 +1,26 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`the HelpCenterButton matches the snapshot 1`] = `
+<button
+  aria-expanded={false}
+  className="sc-htpNat iPmSkf sc-bwzfXH eeZBlJ sc-bdVaJa bsIiYo"
+  onClick={undefined}
```
IreneStr: undefined (see above)

composites/Plugin/Shared/tests/__snapshots__/YoastButtonTest.js.snap
```diff
@@ -2,6 +2,7 @@
 
 exports[`the YoastButton executes callback 1`] = `
 <button
+  aria-expanded={undefined}
```
IreneStr: undefined (see above)

composites/Plugin/Shared/tests/__snapshots__/YoastButtonTest.js.snap
```diff
@@ -14,6 +15,7 @@ exports[`the YoastButton executes callback 1`] = `
 
 exports[`the YoastButton executes callback 2`] = `
 <button
+  aria-expanded={undefined}
```
IreneStr: undefined (see above)

composites/Plugin/Shared/tests/__snapshots__/YoastButtonTest.js.snap
```diff
@@ -26,6 +28,7 @@ exports[`the YoastButton executes callback 2`] = `
 
 exports[`the YoastButton matches the snapshot 1`] = `
 <button
+  aria-expanded={undefined}
```
IreneStr: undefined (see above)

composites/Plugin/Shared/tests/__snapshots__/YoastButtonTest.js.snap
```diff
@@ -26,6 +28,7 @@ exports[`the YoastButton executes callback 2`] = `
...
   className="sc-bwzfXH eeZBlJ sc-bdVaJa IFELj"
   onClick={undefined}
```
IreneStr: undefined (see above)

IreneStr: Not your code, I know, but if you're adding props anyway... ;)

IreneStr:
CR 🚧

composites/AlgoliaSearch/tests/__snapshots__/SearchResultDetailTest.js.snap
afercia: First of all, this test is about the Algolia search "back button" which is unrelated to this PR and I haven't touched the test. This button is not supposed to have an aria-expanded attribute (while the Help Center one does!). It inherits from the YoastButtonBase which now has an aria-expanded attribute based on a `isExpanded` prop. In React, the proper way to _not_ render an attribute is to not pass the related prop, i.e. let it be undefined, so React doesn't render the attribute.
What the snapshots do is not equivalent to the actual rendering.
In a few words, the Back Button must not test the aria-expanded attribute because it doesn't have this attribute set. Instead, the Help Center button does need to have the attribute, if you check its snapshot it has `aria-expanded={false}`.

composites/Plugin/Shared/tests/__snapshots__/YoastButtonTest.js.snap
afercia: same as above

composites/Plugin/Shared/tests/__snapshots__/YoastButtonTest.js.snap
afercia: same as above

composites/Plugin/Shared/tests/__snapshots__/YoastButtonTest.js.snap
afercia: same as above

composites/AlgoliaSearch/tests/__snapshots__/SearchResultDetailTest.js.snap
IreneStr: Ah, thanks for the explanation :)

composites/Plugin/Shared/tests/__snapshots__/HelpCenterButtonTest.js.snap
afercia: Same here, the prop is just not passed.

composites/Plugin/Shared/components/YoastButton.js
afercia: Interesting, although sounds a bit weird to me. So according to the React docs we could  instead of:
`width: PropTypes.number,` also `width: PropTypes.node,` ?

Anyways, I've updated the prop types also for `makeOutboundLink` for consistency, please test with the updated test code sample.

IreneStr:
CR2 👌

IreneStr:
Acceptance 👍

PR #327, opened 2017-10-02
Add callback for used queries

IreneStr:
Adds a prop to the algolia searcher to pass a callback function to retrieve the used search queries.

### DIFF
--- a/composites/AlgoliaSearch/AlgoliaSearcher.js
+++ b/composites/AlgoliaSearch/AlgoliaSearcher.js
@@ -83,6 +83,11 @@ class AlgoliaSearcher extends React.Component {
 		if ( searchString === "" ) {
 			return;
 		}
+		let usedQueries = this.addUsedQuery( searchString );
+
+		if ( this.props.onQueryChange ) {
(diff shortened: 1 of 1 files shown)


composites/AlgoliaSearch/AlgoliaSearcher.js
```diff
@@ -83,6 +83,11 @@ class AlgoliaSearcher extends React.Component {
 		if ( searchString === "" ) {
 			return;
 		}
+		let usedQueries = this.addUsedQuery( searchString );
```
afercia: since now we're using a variable, wouldn't be cleaner to use `usedQueries` below in the setState at line 95 instead of calling `this.addUsedQuery()` again?

afercia:
Then: CR 👍

afercia:
Acceptance 👍 
Tested doing:
```
		<AlgoliaSearcher maxWidth="1280px" onQueryChange={ ( queries ) => {
			console.log( "hello", queries );
		} } enableLiveSearch={ false } />
```
used queries correctly passed.

PR #342, opened 2017-10-05
Improve search results and detail focus management.

### LINKED ISSUE #338: Improve focus management in the KB search results
WHen using a keyboard and activating one of hte KB search results links, the component gets re-rendered:
- the currently focused element (the link) gets removed from the DOM
- focus is lost

In the same way, when in a search detail view, activating the "Go back" button produces a focus loss.

To improve keyboard navigation, fosu should be programmatically set on a proper element.

To evaluate: set focus on the tabpanel itself, which is already focusable and labeled with `aria-labelledby`. Unfortunately, it's part of another component.

afercia:
## Summary

* Improved keyboard focus accessibility in the Knowledge Base search results.

Please refer to the issue for the description of a few technical difficulties to face here. Had to use DOM methods, not particularly happy for that, but couldn't find a better way. Happy to rework this if anyone has better ideas.

## Test instructions

* using the keyboard to navigate, verify focus is not lost
* when going from the search results to a search details focus should go on the detail wrapper
* when going back, focus should go on the result link that was previously clicked
* use a screen reader to check the aria-label on the detail wrapper

Fixes #338

### DIFF
--- a/composites/AlgoliaSearch/AlgoliaSearcher.js
+++ b/composites/AlgoliaSearch/AlgoliaSearcher.js
@@ -312,6 +312,9 @@ class AlgoliaSearcher extends React.Component {
 	getSearchView() {
 		return (
 			<AlgoliaSearchWrapper
+				innerRef={ ( el ) => {
+					this.searchViewWrapper = el;
+				} }
 				maxWidth={ this.props.maxWidth }>
 				{ this.createSearchBar() }
 				{ this.determineSearchResultsView() }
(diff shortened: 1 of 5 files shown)


Xyfi:
Logic for focussing the search result could be moved to SearchResults, but this could be moved to a separate issue.

:+1: CR

boblinthorst:
Acceptance: I checked on both chrome and safari. It follows the behaviour described in the test instructions.

PR #346, opened 2017-10-09
Introduce callbacks for react tabs

afercia:
## Summary

* Needed for https://github.com/Yoast/wordpress-seo/issues/8012

* introduces an utility `triggerDOMEvent` to add custom events and make the react tabs communicate with the jQuery on the plugin

## Test instructions

* See test instructions on the related plugin PR

Fixes https://github.com/Yoast/wordpress-seo/issues/8012


afercia:
Not clear to me how we can do that.
- when the help center is mounted, the react-tabs *panels* are not mounted yet unless we use `forceRenderTabPanel ` see https://github.com/reactjs/react-tabs#forcerendertabpanel-boolean

Triggering an event when the Help Center is mounted in `wordpress-seo` won't help.

- by default only the active tab (in our case the first one) is always rendered
- for the other tabs, their panels are mounted after the react tabs `onSelect`
- we need to communicate to the plugin when the tables are rendered, this is not the same of when the Help Center is rendered

The jQuery part in the plugin needs to calculate the tables size and compare with the parent size.

When the tabs were generated in PHP this was easy, there was just the need to:
- on DOM ready `createScrollableTables()` added some markup around the tables and called `checkScrollableTableSize()`
- when the tabs were clicked, and on window resize and orientation change, we called `checkMultipleScrollableTablesSize()` to check the size of the visible tables 

Now, we need an equivalent to run the same functions and this is tricky because:
- the hidden tab panels are not rendered
- even when a tab panel gets rendered, the tables are injected with `dangerouslySetInnerHTML` so there's no way to detect when a table is rendered

afercia:
> This dom event is too specific for the plugin. I would rather see this even being triggered when the entire help center is mounted in wordpress-seo, so we keep plugin related logic in the plugin as much as possible.

Reminder: as already said, when the help center is mounted, the react tabs are not mounted yet. Only the help center button is rendered. When clicking the button, only in that moment the react tabs are mounted.
```
<HelpCenterButton
	...
</HelpCenterButton>
{ this.state.isExpanded && <HelpCenterPaper minHeight="432px">
	<YoastTabs ...
```

What we need for the jQuery part in the plugin is:
- a callback/event/something when the react tabs are mounted *and their content is rendered*, worth noting the tables gets injected with `dangerouslySetInnerHTML`
- a callback when a tab is selected *and their content is rendered* worth noting the tables gets injected with `dangerouslySetInnerHTML`. for this second case, the react-tabs already have an `onSelect` callback but that alone doesn't guarantee the content injected with `dangerouslySetInnerHTML` is fully rendered.

Needs some discussion on teh best approach :)

afercia:
Removed `triggerDOMevent` and use a callback prop instead.

Testing needs to be done under two different scenarios: when the scrollable tables are in a normal tab (on tab select) and when they're in the initially active tab (on tabs mount).

- yarn link yoast components to the plugin

**On tab select:**
- go in SEO > Titles & Metas 
- click on "Need Help?" to open the help center
- click on "Template explanation"
- verify some HTML elements are prepended and wrapped around the variable tables: They all have class names starting with `yoast-table-scrollable__`
- verify in the responsive view the tables become scrollable

**On tab mount:**
Pass to the `<Tabs/>` component a `defaultIndex={ 3 }` prop. This way, the initially active tab will be the Template explanation one.
- repeat the steps above

Xyfi:
CR 👍 Acceptance 👍

PR #352, opened 2017-10-18
Implement analysis result component

### LINKED ISSUE #191: Implement AnalysisResult component
Using styled components and the Font Awesome eye icon, implement an AnalysisResult component. This component should be a styled `<li>`.

This component is a combination of the IconButton (#190), the Bullet (#189) and a text string. Each AnalysisResult should have a unique, descriptive name/label/id.

IreneStr:
## Summary

* Implements the AnalysisResult component.

## Testing
````js
import React from "react";
import colors from "../../../../style-guide/colors.json";

import AnalysisResult from "../components/AnalysisResult.js";

	min-height: 700px;
	width: 100%;
	background-color: white;
	max-width: 800px;
	margin: 0 auto;
`;

const results = [
	{
		text: "Short text.",
		id: "1",
		rating: "good",
		harMarks: false,
	},
	{
		text: "Longer text. No meta description has been specified. Search engines will display copy from the page instead. No meta description" +
			" has been specified. Search engines will display copy from the page instead.",
		id: "2",
		rating: "OK",
		harMarks: true,
	},
	{
		text: "No meta description has been specified. Search engines will display copy from the page instead.",
		id: "3",
		rating: "bad",
		harMarks: true,
	},
	{
		text: "Not applicable result.",
		id: "4",
		rating: "feedback",
		harMarks: false,
	},
];

/**
 *
 */
export default class ContentAnalysis extends React.Component {
	constructor( props ) {
		super( props );

		this.state = {
			checked: "2",
		};
	}

	handleClick( id ) {
		if ( id === this.state.checked ) {
			this.setState( {
				checked: "-1",
			} );
			return;
		}

		this.setState( {
			checked: id,
		} );
	}

	getColor( rating ) {
		switch ( rating ) {
			case "good":
				return colors.$color_good;
			case "OK":
				return colors.$color_ok;
			case "bad":
				return colors.$color_bad;
			default:
				return colors.$color_score_icon;
		}
	}

	getResults( results ) {
		return results.map( ( result ) => {
			let color = this.getColor( result.rating );

			return <AnalysisResult
				key={ result.id }
				text={ result.text }
				bulletColor={ color }
				hasMarksButton={ result.harMarks }
				ariaLabel="highlight this result in the text"
				pressed={ result.id === this.state.checked }
				buttonId={ result.id }
				onButtonClick={ this.handleClick.bind( this, result.id ) }
			/>;
		} );
	}

	render() {
		return (
			<ContentAnalysisContainer>
				<ul role="list">
					{ this.getResults( results ) }
				</ul>
			</ContentAnalysisContainer>
		);
	}

}
````

[testing code updated by Andrea]

Fixes #191

### CODE REVIEW

File: composites/Plugin/ContentAnalysis/components/AnalysisResult.js [hunk 1]
```diff
@@ -0,0 +1,74 @@
...
+} );
+
+/**
+ * Returns a AnalysisResultBase component.
+ *
+ * @param {object} props Component props.
+ *
+ * @returns {ReactElement} The rendered AnalysisResultBase component.
+ */
+const AnalysisResultBase = styled.li`
```
afercia: As far as I know, we're not documenting styled components. Should this docblock be removed?

composites/Plugin/ContentAnalysis/components/AnalysisResult.js
```diff
@@ -0,0 +1,74 @@
...
+	padding: 4px 4px;
+	display: inline-flex;
+	align-items: center;
+`;
+
+const AnalysisResultText = styled.p`
+	margin: 0 8px;
+`;
+
+export const AnalysisResult = ( props ) => {
```
afercia: Instead, this is a function and should be documented.

composites/Plugin/Shared/components/Icon.js
```diff
@@ -32,3 +33,22 @@ Icon.propTypes = {
 Icon.defaultProps = {
 	size: "16px",
 };
+
+// Safari doesn't expose the ARIA role set on SVG element to browsers.
+export const IconWithAriaLabel = ( props ) => {
```
afercia: This is a function and shoulg have a docblock, like the `Icon` above.

afercia:
CR done 🚧 

A couple functions need docblocks.
In the responsive view, the icons and buttons should be aligned to the top:

screen shot 2017-10-20 at 11 24 52


On the buttons, we're suing the native focus style. While this is clearly perceivable on macOS webkit, it's barely visible on Firefox and other browsers. I'll split this in a separate issue.

screen shot 2017-10-20 at 11 42 57

afercia:
Forgot: the `IconWithAriaLabel` has an aria-label that should be translatable.

afercia:
If we want this component to be managed by the state of a parent component, seems this can be done with no big changes. Completely refactored the testing code, see above.
**Needs to check how data will be passed by the plugin.**

CSS vertical alignments are a bit tricky, they partially depend on the font size and line height which will be different on the plugin and may need adjustments during implementation.

TODO: check translatable strings. Especially the score icons ones:
`ariaLabel={ props.intl.formatMessage( messages.SEOScore ) + " " + props.score }`
we should preferably use replacements. Also, the score "na" needs some special treatment, currently is "SEO score na".

Current rendering:

screen shot 2017-10-20 at 16 56 59

composites/Plugin/ContentAnalysis/components/AnalysisResult.js
```diff
@@ -15,16 +15,23 @@ const messages = defineMessages( {
 	},
 } );
 
+export const AnalysisResultList = styled.ul`
```
IreneStr: This part is already in AnalysisCollapsible.js. This component is meant to be only one analysis result.

composites/Plugin/ContentAnalysis/components/AnalysisResult.js
```diff
@@ -45,15 +52,16 @@ export const AnalysisResult = ( props ) => {
...
-					? <IconButtonToggle
-						onClick={ props.onButtonClick }
-						id={ props.buttonId }
-						icon={ eye }
-						pressed={ props.pressed }
-						ariaLabel={ props.ariaLabel }
-					/>
-					: null
+				props.eyeButtonIsActive &&
+					<span>
```
IreneStr: What is this span doing here?

afercia:
Actually, since the assessment results will be grouped in different groups, the score icons are just decorative now and they don't need a translatable aria-label. Simplified the component and removed react-intl related stuff.

The vertical alignments of the icon etc. may need some adjustments during implementation, since they are affected by the font and line-height.

afercia:
Note for implementation: the buttons need an `ID` and as far as I see the only thing that can be used for this purpose is the assessment identifier, for example: `textTransitionWords`.
Not so ideal because HTML attributes shouldn't be camel-cased, but it doesn't harm.

IreneStr:
CR done 👍

IreneStr:
Acceptance :+1:
I've changed the height of the IconButtonToggle to avoid changes in the height of analysis results when their button is clicked. @afercia CRed that change.

PR #356, opened 2017-10-20
 Add extra test for keywordResultsReducer

IreneStr:
## Summary

Adds an extra test for keywordResultsReducer to test if it overwrites seo results for a single keyword when there are already results for that keyword.

## Testing
* Run `yarn test`.


BramHeijmink:
CR 👍

PR #359, opened 2017-10-24
Implement content analysis component

### LINKED ISSUE #194: Implement ContentAnalysis component
Using styled components, implement a ContentAnalysis component.

This component is a combination the AnalysisHeader (#192) and the AnalysisResults (#191).

Please use react-intl's FormattedMessage to insert the number of problems/improvements/good and to make the texts translatable.

When count = 0 for one of the categories, the header should not be shown.

IreneStr:
## Summary

* Added Content Analysis Component.

* I've decided to make the analysis result categories (errors, improvements etc.) static, although they are not shown when there is no result in that specific category. For now, I don't see any reason to pass categories as a prop, and implementing it that way would make this component way more complex. 
* The bullet is merely decorative. All information needed is communicated through the headings. 
* Discussed header order with @hedgefield: errors, problems, improvements, considerations, good. The collapsibles of errors and problems should be opened by default, the other categories don't.

## Issues still to be solved

* The language notice should only be shown in the readability analysis (not in the SEO analysis). What would be the best way to handle this?
* ~~In the language notice I use the magic number 48px, but that's the only way I get it to work in responsive view:`width: calc(100% - 48px);`~~
* ~~I can't get it to work to bold the language in the language notice and at the same time using a prop to set that language. Assistance is very welcome!~~

## Test instructions

* The content analysis tab is opened by default in the app.
* Margins/paddings are already discussed with @hedgefield. However, suggestions are always welcome, of course. 
* Play around a bit with empty result categories to make sure the respective headings are not shown when there are no results to show.
* Only the 'Errors' and 'Problems' category collapsibles should be opened by default.
* Only one button should be active at any given time (or zero).

Fixes #194

### DIFF
--- a/composites/Plugin/ContentAnalysis/components/AnalysisCollapsible.js
+++ b/composites/Plugin/ContentAnalysis/components/AnalysisCollapsible.js
@@ -68,9 +68,12 @@ const AnalysisList = styled.ul`
 `;
 
 /**
- * @param Component
- * @param headingLevel
- * @returns {XML}
+ * Wrap a component in a header element with a defined heading level.
+ *
+ * @param {ReactElement} Component The component to wrap.
+ * @param {int} headingLevel       The heading level.
+ *
+ * @returns {ReactElement} The wrapped component.
  */
 function wrapInHeading( Component, headingLevel ) {
 	const Heading = `h${ headingLevel }`;
--- a/composites/Plugin/ContentAnalysis/components/AnalysisCollapsible.js
+++ b/composites/Plugin/ContentAnalysis/components/AnalysisCollapsible.js
@@ -11,7 +11,7 @@ import { IconButton } from "../../Shared/components/Button";
  * Container for the Collapsible header and its content.
  */
 const AnalysisHeaderContainer = styled.div`
-	margin: 8px 0;
+	margin: 4px 0 8px 0;
 	background-color: ${ colors.$color_white };
 `;
 
@@ -25,21 +25,24 @@ const AnalysisHeaderButton = styled( IconButton )`
 	border-color: transparent;
 	border-radius: 0;
 	outline: none;
+	justify-content: left;
 	box-shadow: none;
 	// When clicking, the button text disappears in Safari 10 because of color: activebuttontext.
 	color: ${ colors.$color_blue };
 
 	:hover {
 		border-color: transparent;
+		color: ${ colors.$color_blue };
 	}
 
 	:active {
 		box-shadow: none;
 		background-color: ${ colors.$color_white };
+		color: ${ colors.$color_blue };
 	}
 
 	svg {
-		margin: 0 10px;
+		margin: 0 8px;
 		width: 20px;
 		height: 20px;
 	}
@@ -61,7 +64,7 @@ const AnalysisTitle = styled.span`
 const AnalysisList = styled.ul`
 	margin: 0;
 	list-style: none;
-	padding: 8px 16px;
+	padding: 0 16px 0 13px;
 `;
 
 /**
--- a/composites/Plugin/ContentAnalysis/components/AnalysisCollapsible.js
+++ b/composites/Plugin/ContentAnalysis/components/AnalysisCollapsible.js
@@ -25,7 +25,7 @@ const AnalysisHeaderButton = styled( IconButton )`
 	border-color: transparent;
 	border-radius: 0;
 	outline: none;
-	justify-content: left;
+	justify-content: flex-start;
 	box-shadow: none;
 	// When clicking, the button text disappears in Safari 10 because of color: activebuttontext.
 	color: ${ colors.$color_blue };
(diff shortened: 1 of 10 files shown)


Xyfi:
Applied some codestyle fixes and applied a little fix to make the language bold in the language notice.

composites/Plugin/ContentAnalysis/components/ContentAnalysis.js
```diff
@@ -32,7 +32,7 @@ const ChangeLanguageLink = makeOutboundLink( styled.a`
 const messages = defineMessages( {
 	languageNotice: {
 		id: "content-analysis.language-notice",
-		defaultMessage: "Your site language is set to %s.",
+		defaultMessage: "Your site language is set to {language}.",
```
IreneStr: Please remove this message entry. This extra step is not necessary.

IreneStr:
CR for Alexander's changes: done 👍

composites/Plugin/ContentAnalysis/components/ContentAnalysis.js
```diff
@@ -1,17 +1,218 @@
...
+					: null }
+				{ considerationsResults.length > 0
+					? <AnalysisCollapsible title={ this.props.intl.formatMessage( messages.considerationsHeader ) }>
+						{ this.getResults( considerationsResults ) }
+					</AnalysisCollapsible>
+					: null }
+				{ goodResults.length > 0
+					? <AnalysisCollapsible title={this.props.intl.formatMessage( messages.goodHeader ) }>
+						{ this.getResults( goodResults ) }
+					</AnalysisCollapsible>
```
afercia: If I'm not wrong, instead of these 5 ternaries:
condition ? component : null
we could simplify and just do:
condition && component

afercia:
Tested also on IE 11 and Edge, works.

However, for acceptance, I guess this component should solve one of the original issues see https://github.com/Yoast/wordpress-seo/issues/440

In fact, one of the purposes of splitting the results in different groups was to use headings to identify the results groups. This way, the score icon would become just decorative, each group would be identified by a proper heading (h2, or h3, whatever). 

The `AnalysisCollapsible` is just a button though, so the groups have no headings. Maybe we could easily modify the AnalysisCollapsible to optionally output a heading.

Native heading elements are preferable, something like:

`<h2><button /></h2>`

alternatively, some ARIA attributes could help, e.g. `role="heading" aria-level="2"` but I'd recommend native headings.

afercia:
Testing the headings 🙂 

with headings

composites/Plugin/ContentAnalysis/components/AnalysisCollapsible.js
```diff
@@ -68,6 +68,28 @@ const AnalysisList = styled.ul`
 `;
 
 /**
+ * @param Component
```
IreneStr: Please add docs

composites/Plugin/ContentAnalysis/components/AnalysisCollapsible.js
```diff
@@ -78,33 +100,18 @@ const AnalysisList = styled.ul`
 export const AnalysisCollapsibleStateless = ( props ) => {
 	let title = props.title;
 	let count = getChildrenCount( props.children );
-	const Heading = `h${ props.headingLevel }`;
 
-	const StyledHeading = styled( Heading )`
-		margin: 0;
-		font-weight: normal;
-    `;
+	const Button = props.hasHeading ? wrapInHeading( AnalysisHeaderButton, 2 ) : AnalysisHeaderButton;
```
IreneStr: This should not be a hardcoded 2

composites/Plugin/ContentAnalysis/components/AnalysisCollapsible.js
```diff
@@ -78,33 +100,18 @@ const AnalysisList = styled.ul`
...
-				>
-					<AnalysisTitle>{ `${ title } (${ count })` }</AnalysisTitle>
-				</AnalysisHeaderButton>
-			}
+			<Button
+				aria-expanded={ props.isOpen }
+				onClick={ props.onToggle }
+				icon={ props.isOpen ? angleUp : angleDown }
+				iconColor={ colors.$color_grey_dark } >
+				<AnalysisTitle>{ `${ title } (${ count })` }</AnalysisTitle>
```
IreneStr: spaces

afercia:
Much cleaner with `wrapInHeading()` 🎉  

Pushed a few coding standards:
- spaces
- as far as I know, we don't document styled-components
- params names and descriptions should always be aligned, this is something we're not doing consistently through the codebase and we should try to improve it

afercia:
Note: maybe we should consider to move to a separate file utilities and HOCs like `wrapInHeading()`. For example, `makeOutboundLink()` is already in its own file under `utils/`.

Also, maybe `wrapInHeading()` should not contain styling, instead styling should be applied on a case by case basis.

Material for a separate issue though 🙂

IreneStr:
@afercia Sounds like a good idea. Please create a separate issue :)

PR #364, opened 2017-10-27
Removed package-lock.json from .gitignore

Xyfi:
Should not have been added in the first place, due to miscommunication. NPM documentation clearly states `package-lock.json` should be added to VC.


atimmer:
We don't want to maintain both a `yarn.lock` and a `package-lock.json`.

PR #369, opened 2017-11-06
Make the Wizard previous and next buttons focusable again

### LINKED ISSUE #368: Wizard previous/next buttons are not focusable
On the wizard, the previous and next buttons are now not focusable and can't be used with a keyboard.

screen shot 2017-11-06 at 10 38 35

Checking the wizard on the plugin current version, these buttons have a tabindex=0.
However, in the yoast-components, they get a tabindex=-1.

I guess something has changed in `material-ui` and to my understanding, `disableKeyboardFocus` was previously used just for styling purposes. Instead, now it is used also to control the buttons focusability. See https://github.com/Yoast/yoast-components/pull/70 where `disableKeyboardFocus` was used for styling purposes.

I think we should just remove `disableKeyboardFocus: true`.

Aside: as far as I see, `material-ui` does lots of things with focusability of elements, often overriding the native behavior and rebuilding it programmatically. I'm not sure I like so much this approach :(

afercia:
## Summary

* Made the Onboarding Wizard previous and next buttons focusable again

* Removes the `disableKeyboardFocus` from the buttons. Seems this prop was initially used in material-ui just for styling purposes, instead now it's also responsible for setting the tabindex attribute.

## Test instructions

- on the develop branch, use the keyboard to Tab to the buttons and see they're not focusable
- see they have a `tabindex="-1"` attribute
- switch to this branch
- check the buttons are now focusable
- check they have a `tabindex="0"` attribute

Aside note: as mentioned in the issue, `material-ui` does lots of things to control elements focusability programmatically. Not sure I really like this approach. Keeping controls behaviors as native as possible is always a better choice.

Fixes #368

### DIFF
--- a/composites/OnboardingWizard/OnboardingWizard.js
+++ b/composites/OnboardingWizard/OnboardingWizard.js
@@ -278,7 +278,6 @@ class OnboardingWizard extends React.Component {
 				onClick: this.setPreviousStep,
 				disableFocusRipple: true,
 				disableTouchRipple: true,
-				disableKeyboardFocus: true,
 				icon: <ArrowBackwardIcon viewBox="0 0 28 28" />,
 			}, step, "yoast-wizard--button yoast-wizard--button__previous" );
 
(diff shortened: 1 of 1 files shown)


Xyfi:
CR :+1:

Xyfi:
Acceptance 👍

PR #386, opened 2017-11-22
Avoid collapsible header button full re-render.

### LINKED ISSUE #385: Content Analysis Collapsible a11y improvements
Splitting this out from https://github.com/Yoast/wordpress-seo/issues/8310. For more details, please see the comments thre.

Currently, when the AnalysisCollapsible button is wrapped in the optional heading, the button is fully re-rendered every time it gets clicked. This produces two behaviors that impact negatively accessibility:
- since the button gets fully re-rendered, there's a focus loss: modern browsers try to be smart and keep focus in place, but if you tab you will see focus is not on the button. In IE 11 focus fallbacks to the document root
- since the button loses focus, screen readers don't announce the aria-expanded state correctly

To my understanding, this is expected in React because the parent heading gets re-rendered and thus the child button gets re-rendered too.
Worth noting every time the `isOpen` state changes the `AnalysisCollapsibleStateless` gets re-rendered, and `wrapInHeading()` runs every time.

afercia:
## Summary

* Avoid to fully re-render the collapsible header to improve accessibility

* moved the evaluation of the "wrapping" to avoid to run `wrapInHeading()` in the stateless component, which is rendered every time a button gets clicked

## Test instructions

- before switching to this branch, go in the Content Analysis tab
- using Safari and VoiceOver, focus one ot the results heading
- using Control-Option-Spacebar, activate the button in the heading
- do it on different headings, multiple times
- observe how the `aria-expanded` change is not announced correctly and there's a focus loss, sometimes it's very evident focus goes on a different element
- switch to this branch
- repeat the steps above
- observe there's no focus loss and the `aria-expanded` change is announced correctly

You can also observe the Chrome inspector and see how the heading gets fully re-rendered before this changes, while in this PR it just gets updated.

Fixes #385

### DIFF
--- a/composites/Plugin/ContentAnalysis/components/AnalysisCollapsible.js
+++ b/composites/Plugin/ContentAnalysis/components/AnalysisCollapsible.js
@@ -90,17 +90,17 @@ export const AnalysisCollapsibleStateless = ( props ) => {
 	let title = props.title;
 	let count = getChildrenCount( props.children );
 
-	const MaybeWrappedButton = props.element;
+	const Button = props.element;
 
 	return (
 		<AnalysisHeaderContainer>
-			<MaybeWrappedButton
+			<Button
 				aria-expanded={ props.isOpen }
 				onClick={ props.onToggle }
 				icon={ props.isOpen ? angleUp : angleDown }
 				iconColor={ colors.$color_grey_dark } >
 				<AnalysisTitle>{ `${ title } (${ count })` }</AnalysisTitle>
-			</MaybeWrappedButton>
+			</Button>
 			{
 				props.isOpen && props.children &&
 					<AnalysisList role="list">{ props.children }</AnalysisList>
--- a/composites/Plugin/ContentAnalysis/components/AnalysisCollapsible.js
+++ b/composites/Plugin/ContentAnalysis/components/AnalysisCollapsible.js
@@ -90,21 +90,20 @@ export const AnalysisCollapsibleStateless = ( props ) => {
 	let title = props.title;
 	let count = getChildrenCount( props.children );
 
-	const Button = props.hasHeading ? wrapInHeading( AnalysisHeaderButton, props.headingLevel ) : AnalysisHeaderButton;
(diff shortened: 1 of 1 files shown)


afercia:
Note: moving the "wrapping" evaluation to the stateful component seems the easier fix to me. Will leave technical considerations to @Xyfi expertise 🙂

composites/Plugin/ContentAnalysis/components/AnalysisCollapsible.js
```diff
@@ -90,21 +90,20 @@ export const AnalysisCollapsibleStateless = ( props ) => {
 	let title = props.title;
 	let count = getChildrenCount( props.children );
 
-	const Button = props.hasHeading ? wrapInHeading( AnalysisHeaderButton, props.headingLevel ) : AnalysisHeaderButton;
+	const MaybeWrappedButton = props.element;
```
Xyfi: I would just keep calling it Button. It maybe wrapped in a heading, but it is still a button.

composites/Plugin/ContentAnalysis/components/AnalysisCollapsible.js
```diff
@@ -120,6 +119,7 @@ AnalysisCollapsibleStateless.propTypes = {
 		PropTypes.arrayOf( PropTypes.node ),
 		PropTypes.node,
 	] ),
+	element: PropTypes.func,
```
Xyfi: Can be changed to PropTypes.element

Xyfi:
I confirmed it works, but I would like to look into it again tomorrow to see why this works exactly.

composites/Plugin/ContentAnalysis/components/AnalysisCollapsible.js
afercia: Changing to element, I get:
Warning: Failed prop type: Invalid prop `element` of type `function` supplied to `AnalysisCollapsibleStateless`, expected a single ReactElement.

Xyfi:
CR: +1:

PR #391, opened 2017-11-29
Automatically open relevant collapsible

IreneStr:
## Summary

* Expands the AnalysisCollapsible(s) that are currently the most urgent in the content analysis. For example: if there are no problems, but there are results in the categories 'needs improvement' and 'good', the 'needs improvement' AnalysisCollapsible is expanded by default. 

## Test instructions

* Link this branch to your local Wordpress environment with the wordpress-seo release/5.9 branch. 
* Play around with empty result categories. The most urgent category should always be expanded. The order is: errors, problems, needs improvements, considerations, good. For example: Write a post with no 'problems', but with 'needs improvements'. Make sure 'needs improvements' is expanded by default.
* **Not fixed in this PR:** Write a post with one 'problem' and one or more 'needs improvement's. Solve the problem. The 'needs improvement' should expand (and the 'problem' should disappear, of course).

Partial fix of: https://github.com/Yoast/wordpress-seo/issues/8392. See comment on that issue for more context.


composites/Plugin/ContentAnalysis/components/ContentAnalysis.js
```diff
@@ -219,6 +219,8 @@ class ContentAnalysis extends React.Component {
 					<AnalysisCollapsible
 						hasHeading={ true }
 						headingLevel={ headingLevel }
+						// eslint-disable-next-line no-inline-comments
+						initialIsOpen={ problemsResults.length <= 0 } /* Only open when there are no problems to be solved */
```
moorscode: Could these be less then zero?

If you create a variable above this like `problemsFound = problemResults.length` you would not need a comment either. (Also applying this for the other Results conditions below of course).

Which would make it:
```
initialIsOpen={ problemsFound == 0 }
```

IreneStr:
Decided to not merge this until the underlying issues in https://github.com/Yoast/wordpress-seo/issues/8392 have been fixed.

Temporary fix: https://github.com/Yoast/yoast-components/pull/391

IreneStr:
CR done 👍

IreneStr:
Acceptance 🚧  
If I close 'problems', fix the remaining problems (--> improvements heading is expanded), then cause a problem, the problems heading is expanded. As far I understand it, that is something we'd want to avoid, because we'd like to respect the user's choice to close 'problems'. Am I correct @atimmer?

IreneStr:
Found the cause of the double-click issue: the initial state of all collapsibles is `null`. When 'problems' is clicked, this becomes `! null = true`, but the collapsible is already opened, so it seems like it's not responding.

jcomack:
CR 2 done 👍

IreneStr:
Acceptance :+1: Second acceptance test by Jimmy on its way

jcomack:
Acceptance done 👍

PR #409, opened 2017-12-11
Add prop to disable and hide marking buttons

### LINKED ISSUE #408: Content analysis: marks buttons missing tooltips
Couple things:

1
in the non-react version, the marks buttons used to have tooltips

2
the aria-label was dynamic, while on the React version is always `Highlight this result in the text` regardless of the mark button state

Screenshots: how the tooltips used to be:

screen shot 2017-12-11 at 10 04 05

screen shot 2017-12-11 at 10 04 12

screen shot 2017-12-11 at 10 04 30

See also https://github.com/Yoast/wordpress-seo/issues/8442

IreneStr:
## Summary

* Added a prop to the `ContentAnalysis` to disable or hide the eye buttons.

* I've reset the cursor to `default` rather than `pointer` on hovering a disabled button. This is consistent with what WP is doing. 
* I've changed the icon color of the disabled button to #ddd ($color_grey) instead of #e6e6e6 (which is not in the colors.json). #ddd is a little darker, but the background is darker as well,
 because now there is a button background instead of the white meta box background. The color contrast is 1.27 instead of 1.25 🎉 .
* The design of the disabled button has been approved by @hedgefield.

## Test instructions

* Check out this branch.
* `yarn install`,`yarn start`
* Go to `localhost:3333`
* Change the marksButtonStatus in `app/ContentAnalysisWrapper.js` to 'hidden' and 'disabled' and check whether the buttons are hidden or disabled.

Part of the fix for https://github.com/Yoast/wordpress-seo/issues/8442
Fixes #408

### DIFF
--- a/composites/Plugin/ContentAnalysis/components/AnalysisResult.js
+++ b/composites/Plugin/ContentAnalysis/components/AnalysisResult.js
@@ -26,10 +26,14 @@ const AnalysisResultText = styled.p`
 	flex: 1 1 auto;
 `;
 
+/**
+ * Determines whether the buttons should be hidden.
+ *
+ * @param {Object} props The component's props.
+ * @returns {boolean} True if buttons should be hidden.
+ */
 let areButtonsHidden = function( props ) {
-	if ( props.marksButtonStatus === "hidden" ) {
-		return true;
-	}
+	return props.marksButtonStatus === "hidden";
 };
 
 /**
--- a/composites/Plugin/ContentAnalysis/components/AnalysisResult.js
+++ b/composites/Plugin/ContentAnalysis/components/AnalysisResult.js
@@ -46,6 +46,7 @@ export const AnalysisResult = ( props ) => {
 				props.hasMarksButton &&
 					<IconButtonToggle
 						buttonsDisabled={ props.buttonsDisabled }
+						className={ props.markButtonClassName }
 						onClick={ props.onButtonClick }
 						id={ props.buttonId }
 						icon={ eye }
@@ -66,6 +67,7 @@ AnalysisResult.propTypes = {
 	ariaLabel: PropTypes.string.isRequired,
 	onButtonClick: PropTypes.func.isRequired,
 	buttonsDisabled: PropTypes.bool,
+	markButtonClassName: PropTypes.string,
 };
 
 AnalysisResult.defaultProps = {
--- a/composites/Plugin/ContentAnalysis/components/AnalysisResult.js
+++ b/composites/Plugin/ContentAnalysis/components/AnalysisResult.js
@@ -26,6 +26,12 @@ const AnalysisResultText = styled.p`
 	flex: 1 1 auto;
 `;
 
+let areButtonsHidden = function( props ) {
+	if ( props.buttonStatus === "hidden" ) {
+		return true;
+	}
+};
+
 /**
  * Returns an AnalysisResult component.
  *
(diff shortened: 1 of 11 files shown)


composites/Plugin/ContentAnalysis/components/AnalysisResult.js
```diff
@@ -26,6 +26,12 @@ const AnalysisResultText = styled.p`
 	flex: 1 1 auto;
 `;
 
+let areButtonsHidden = function( props ) {
```
andizer: Missing docs?

andizer: This method is returning `true` when the condition in the if is met. But there is no default.

composites/Plugin/ContentAnalysis/components/AnalysisResult.js
```diff
@@ -26,6 +26,12 @@ const AnalysisResultText = styled.p`
...
+	if ( props.marksButtonStatus === "hidden" ) {
```
andizer: You might consider: `return ( props.marksButtonStatus === "hidden" );` instead of the if.

composites/Plugin/ContentAnalysis/components/ContentAnalysis.js
```diff
@@ -132,15 +140,26 @@ class ContentAnalysis extends React.Component {
 	getResults( results ) {
 		return results.map( ( result ) => {
 			let color = this.getColor( result.rating );
+			let isPressed = result.id === this.state.checked;
+			let ariaLabel = "";
+			if ( this.props.marksButtonStatus === "disabled" ) {
```
andizer: I would place most of this if-else statement outside of the mapping:
```
if ( this.props.marksButtonStatus === "disabled" ) {
   defaultAriaLabel = this.props.intl.formatMessage( messages.disabledButton );
} else {
   defaultAriaLabel = this.props.intl.formatMessage( messages.highlight );
}
```

If you put this before the return then you can do the following in the map callback:
```
let ariaLabel = defaultAreaLabel;
if ( this.props.marksButtonStatus === "disabled" ) {
    ariaLabel = this.props.intl.formatMessage( messages.noHighlight );
}
```

composites/Plugin/Shared/components/IconButtonToggle.js
```diff
@@ -21,10 +21,24 @@ const ChangingIconButtonBase = styled.button`
...
+		box-shadow: none;
+		border: none;
+		cursor: default;
 	}
 `;
 
+let areButtonsDisabled = function( props ) {
+	if ( props.marksButtonStatus === "disabled" ) {
+		return true;
+	} else if ( props.marksButtonStatus === "enabled" ) {
```
andizer: Instead of `else if` you simply can do `if....`

composites/Plugin/Shared/components/IconButtonToggle.js
```diff
@@ -21,10 +21,24 @@ const ChangingIconButtonBase = styled.button`
...
+	}
+	&:disabled {
+		background-color: ${ props => props.unpressedBackground };
+		box-shadow: none;
+		border: none;
+		cursor: default;
 	}
 `;
 
+let areButtonsDisabled = function( props ) {
```
andizer: Missing docs

andizer: You might consider returning a default bool. Because now this method returns two different types: `void` and `bool`

composites/Plugin/ContentAnalysis/components/ContentAnalysis.js
IreneStr: Those three ifs are too intertwined to do that

andizer:
Acceptance done. Will merge this pull when acceptance test for plugin implementation is done.

PR #416, opened 2018-01-22
Styled section as styled component

afercia:
For the "Insights" section on Premium, needs:
https://github.com/Yoast/wordpress-seo-premium/pull/1567
https://github.com/Yoast/yoast-premium-components/pull/43

- extract the heading in a separate component, with test
- makes the Section heading optional, rename props
- introduces a StyledSection component with a Heading an optional icon
- introduces a SvgIcon component

To check: are the Section and StyledSection currently used in other components? Haven't found other usages, any help in double checking very welcome.

To test the standalone version, use this code in App.js. Test changing props: heading level, heading color, icon, icon color, etc.
```
import React from "react";
import styled from "styled-components";

import StyledSection from "./forms/StyledSection/StyledSection";
import Section from "./forms/Section";

const Container = styled.div`
	max-width: 800px;
	margin: 0 auto;
`;

class App extends React.Component {

	constructor() {
		super();
	}

	render() {
		return (
			<Container>
				<StyledSection
					headingText="Insights. Hello, this is a Styled Section heading."
					headingColor="red"
					headingLevel={ 2 }
					headingIcon="file-text"
					headingIconColor="blue"
					headingIconSize="16"
				>
					<p>Keyword Suggestions</p>
					<p>Hello, this is additional content</p>
				</StyledSection>
				<br />
				<Section
					headingText="Hello, this is a normal Section heading"
					headingLevel={ 3 }
				>
					<p>Normal Section content</p>
				</Section>
			</Container>
		);
	}
}

export default App;

```

Testing the Premium "Insights" section is a bit more complicated. What I've done:
- switch Premium and `yoast-premium-components` to the branches related to the PRs above
- on `yoast-premium-components`, yarn link `yoast-components`
- on premium, inside the `premium/` directory, yarn link `yoast-components`
- on premium, inside the `premium/` directory, build the JS

This should be enough to see the "Insights" section updated.

Note: I've tested the standalone version on Windows, IE 11 and Edge, everything looks fine. I can't test Premium on Windows because Git on Windows hates me.

Fixes #415

PR #421, opened 2018-02-01
Remove svg files, the Icon component, and svg-react-loader

### LINKED ISSUE #420: Replace svg files
I propose to remove the svg files from `yoast-components`, because they are converted to react components anyway by `svg-react-loader`. 

In PR #416 @afercia created a new `SvgIcon` component, which can be used from now on instead of importing actual .svg files.

afercia:
## Summary

* REmoved the Icon component, replaced by the new SvgIcon React component

- this PR is branched out from `stories/415-styled-styled-section`
- removes the Icon component
- removes the svg files from `style-guide/svg`
- uses StyledSection where possible
- adapt other components, e.g. IconButton, IconButtonToggle, to the new SvgIcon component introduced in `stories/415-styled-styled-section`
- removes svg-react-loader

I've looked for all the imports of icons in the other repos and tried to address them. Please help me check no other usages of the SVG files and hte `<Icon>` component have survived and nothing breaks 🙂 
@Xyfi please check the `svg-react-loader` stuff, I have no idea what I've done there 😆 

## Test instructions

- check all the components with icons work correctly in the standalone version
- switch both free and premium to their `stories/415-styled-styled-section` branches if they're not merged yet
- yarn link this branch to free and premium, build the assets, and check all the icons
- on Premium, you will get an error when building the CSS because `StyledSection.scss` was removed, this is covered on free in https://github.com/Yoast/wordpress-seo/pull/8812 I guess your best option is to temporarily remove that import manually

Fixes #420


composites/Plugin/Shared/components/HelpCenterButton.js
```diff
@@ -17,9 +16,9 @@ import { Icon } from "../../Shared/components/Icon";
 const HelpCenterButtonBase = ( props ) => {
 	return (
 		<YoastButton { ...props }>
-			<Icon icon={ questionCircle } color={ props.iconColor } />
+			<SvgIcon icon="questionCircle" color={ props.iconColor } />
```
Xyfi: I believe this should be `"question-circle"`.

composites/Plugin/Shared/components/SvgIcon.js
```diff
@@ -0,0 +1,96 @@
...
+			case "search":
+				path = "M1216 832q0-185-131.5-316.5t-316.5-131.5-316.5 131.5-131.5 316.5 131.5 316.5 316.5 131.5 316.5-131.5 131.5-316.5zm512 832q0 52-38 90t-90 38q-54 0-90-38l-343-342q-179 124-399 124-143 0-273.5-55.5t-225-150-150-225-55.5-273.5 55.5-273.5 150-225 225-150 273.5-55.5 273.5 55.5 225 150 150 225 55.5 273.5q0 220-124 399l343 343q37 37 37 90z";
+				break;
+			case "times":
+				path = "M1490 1322q0 40-28 68l-136 136q-28 28-68 28t-68-28l-294-294-294 294q-28 28-68 28t-68-28l-136-136q-28-28-28-68t28-68l294-294-294-294q-28-28-28-68t28-68l136-136q28-28 68-28t68 28l294 294 294-294q28-28 68-28t68 28l136 136q28 28 28 68t-28 68l-294 294 294 294q28 28 28 68z";
+				break;
+		}
+		/* eslint-enable max-len */
+
+		if ( ! path ) {
```
Xyfi: Maybe add a `console.warn` for easier debugging?

composites/Plugin/Shared/components/Icon.js
```diff
@@ -1,35 +0,0 @@
-import React from "react";
```
Xyfi: I don't think this file has to be removed, because it can still be used to render svg react components. We could still have more complex react components (that use more than just a path) that are just plain react components that are passed to this component.

afercia: OK, however the Icon component had also a IconTest.js, which required the edit.svg file.

afercia: And should we keep `svg-react-loader` then?

Xyfi: No, in case we keep using `Icon.js`, we start using SVG's are react components and not as svg files:

```js
// React component
const EditIcon = ( props ) => {
    return(
        <svg { ...props } >{ "..." }</svg>
    )
}

**Comment by afercia** [comment]
Note: I guess this should be rebased interactively after #416 gets merged.

**Comment by afercia** [comment]
Instead of building new components for the SVGs, I've adjusted the existing `YoastLogo..js` and `YoastSeoIcon.js` to work with the `<Icon />` component, thus wrapper in a styled components. Building new ones would have led to duplication, as the new components would have been very similar to the existing ones.

Adjusted the usage of the icons in the Onboarding Wizard.

To see hte new icons in action, use this code in `App.js` or yarn-link yoast-components to the plugin ((the MessageBox gets displayed when, for some reasons, the configuration can't be loaded or is empty)


```
import React from "react";
import { IntlProvider } from "react-intl";

import Icon from "./composites/Plugin/Shared/components/Icon";
import YoastLogo from "./composites/basic/YoastLogo"
import YoastSeoIcon from "./composites/basic/YoastSeoIcon"

const Container = styled.div`
	width: 800px;
	margin: 0 auto;
	text-align: center;
`;

class App extends React.Component {

	constructor() {
		super();
	}

	render() {
		return (
			<IntlProvider locale="en">
				<Container>
					<Icon icon={ YoastLogo } width="200px" height="93px" />
					<br />
					<Icon icon={ YoastSeoIcon } color="red" width="100px" height="100px" />
				</Container>
			</IntlProvider>
		);
	}
}

export default App;

```

afercia:
P.S. I haven't found any other usage of `YoastLogo.js` and `YoastSeoIcon.js` but worth double checking.

afercia:
Note: the `color` prop doesn't have any effect when the SVG provides its own fill colors for the shape. It will work for monochromatic icons that don't have a `fill` attribute in the SVG markup.

Xyfi:
CR :+1:

igorschoester:
I added StyledSection tests base on @afercia 's test `App.js` code in this [PR](https://github.com/Yoast/yoast-components/pull/416). A side effect of this is that a console warning is shown on the test with a non-existing icon 😅 

Acceptance 👍

PR #427, opened 2018-02-20
Fix default props in svg icon component

Xyfi:
## Summary

* Changed `SvgIcon`'s default prop `size` from `16` to `16px` and correctly implemented default props.


IreneStr:
CR and acceptance done 👍

PR #431, opened 2018-02-28
Change initial state to Open

### LINKED ISSUE #430: Expand analysis headings by default
As per https://github.com/Yoast/wordpress-seo/issues/8882 - we don't want to keep the `green circles` from being enjoyed by the users of the plugin.

Thus expanding all analysis headings by default is approach we want to take.
This will still allow a user to collapse a certain heading.

moorscode:
Remove conditional isOpen state variables
Add additional test for collapsed state snapshot
Update snapshots

## Summary

* Changes the default state for Analysis headers to Collapsed instead of conditionally based upon the most important issues.

* Removed the value from the components when they match the default.

## Test instructions

* Open the Analysis and see the headings being expanded.
* The headings should still be collapsible.

Fixes #430


igorschoester:
Acceptance 👍

PR #439, opened 2018-03-14
Reactify checkbox

### LINKED ISSUE #435: Reactify cornerstone content checkbox
Please make sure this component can be reused for other checkboxes in the future, by passing the label as a prop. 
schermafbeelding_2018-03-13_om_12_02_57

The component should be able to keep state (checked or unchecked)

maartenleenders:
## Summary

* Created a Checkbox React component, including a label.

* WordPress styling will overwrite styled.components, but we decided to cross that bridge when we get there.

## Test instructions

* Checkout branch, **yarn start**, and navigate to http://localhost:3333/
* In Checkbox, there is the most generic form of a react checkbox and label.
* Clicking the label and the box should (un)check the box, and in the console it should log the value of the checkbox.
* In `App.js` you can toy around with the props that are passed to it.

Fixes #435

### DIFF
--- a/composites/Plugin/Shared/components/Checkbox.js
+++ b/composites/Plugin/Shared/components/Checkbox.js
@@ -7,7 +7,7 @@ const YoastCheckbox = styled.input`
 `;
 
 /**
- * Creates a React Checkbox and Label component.
+ * Creates a React Checkbox which consists of a checkbox and a label.
  */
 class Checkbox extends React.Component {
 	constructor( props ) {
@@ -31,7 +31,7 @@ class Checkbox extends React.Component {
 	/**
 	 * Renders a checkbox and a label.
 	 *
-	 * @returns {ReactElement} The Checkbox react component including label.
+	 * @returns {ReactElement} The Checkbox react component including its label.
 	 */
 	render() {
 		return(
--- a/composites/Plugin/Shared/components/Checkbox.js
+++ b/composites/Plugin/Shared/components/Checkbox.js
@@ -25,8 +25,7 @@ class Checkbox extends React.Component {
 	 * @returns {void} Fires the onChange function passed in the props.
 	 */
 	handleChange( event ) {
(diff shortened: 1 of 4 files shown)


composites/Plugin/Shared/components/Checkbox.js
```diff
@@ -0,0 +1,65 @@
...
+		PropTypes.string,
+		PropTypes.array,
+	] ),
+	checked: PropTypes.bool,
+
+};
+
+Checkbox.defaultProps = {
+	checked: false,
+	label: "",
```
afercia: While _not_passing a label would be very evident because only the checkbox would be visible, it would also print out an empty label element e.g. `<label for="exampleCheckbox"></label>`. We should avoid this and I'd suggest to make the label prop required instead of setting empty string as default value.

composites/Plugin/Shared/components/Checkbox.js
```diff
@@ -0,0 +1,65 @@
+import React from "react";
+import styled from "styled-components";
+import PropTypes from "prop-types";
+
+const YoastCheckbox = styled.input`
+	margin-right: 0.5em;
+`;
+
+/**
+ * Creates a React Checkbox and Label component.
```
IreneStr: There is no label component

composites/Plugin/Shared/tests/CheckboxTest.js
```diff
@@ -0,0 +1,47 @@
...
+	);
+
+	let tree = component.toJSON();
+	expect( tree ).toMatchSnapshot();
+
+	tree[ 0 ].props.onChange( event );
+
+	tree = component.toJSON();
+	expect( tree ).toMatchSnapshot();
+} );
```
IreneStr: Include a test that also includes a state change?

Xyfi:
CR & Acceptance :+1:

PR #441, opened 2018-03-14
Use React Modal for modals [WIP]

afercia:
Work in progress for the modals.

Fixes https://github.com/Yoast/wordpress-seo/issues/5781


afercia:
Thanks @igorschoester so I've started from http://reactcommunity.org/react-modal/testing/ but both methods mentioned there didn't work for me (maybe I was doing it wrong, not sure)
Then I've read a few things including the issue you've linked but still... in the second test, if you dump this after the `modal` variable is set:
`console.log( modal.find( ".yoast-modal__content" ).length );`
it will return `3` 😬  
Same weirdness when trying to get the elements below for example change
`expect( modal.find( "button.yoast-modal__button-close-icon" ).getDOMNode().getAttribute( "aria-label" ) ).toBe( "Close X" );`

to
`expect( modal.find( ".yoast-modal__button-close-icon" ).getDOMNode().getAttribute( "aria-label" ) ).toBe( "Close X" );`

and I get
`Method “getDOMNode” is only meant to be run on a single node. 2 found instead.`

weird because I can't understand how removing `button` can make a difference.

afercia:
> it will return 3

removing `isOpen={ true }` hten it returns 2, I guess the test is just mounting the modal multiple times.

afercia:
Okay I'm getting closer, maybe :) Lots of wrapper here. the test wrapper, the YoastModal (Styled Component), the BaseYoastModal, and the react-modal... dumping
`console.log( wrapper.find( BaseYoastModal ).debug() );`
I see this:

screen shot 2018-03-30 at 10 12 53

ant actually there are 3 `.yoast-modal__content`

composites/Plugin/Shared/components/YoastModal.js
```diff
@@ -0,0 +1,155 @@
...
+import colors from "../../../../style-guide/colors.json";
+import SvgIcon from "./SvgIcon";
+
+const StyledHeading = styled.h1`
+	float: left;
+	margin: -4px 0 2rem;
+	font-size: 1rem;
+`;
+
+const StyledButton = styled.button`
```
Xyfi: Could this be named a bit more specific to the function it has, since it is specifically for the x-cross button?

composites/Plugin/Shared/components/YoastModal.js
```diff
@@ -0,0 +1,155 @@
...
+				{
+					this.props.heading &&
+						<StyledHeading className="yoast-modal__title">{ this.props.heading }</StyledHeading>
+				}
+				{
+					this.props.closeIconButton &&
+						<StyledButton
+							onClick={ this.props.onClose }
+							className={ `${ this.props.xCloseButtonClassName } yoast-modal__button-close-icon` }
+							aria-label={ this.props.closeIconButton }
```
Xyfi: shouldn't this be `ariaLabel`?

composites/Plugin/Shared/components/YoastModal.js
Xyfi: Is there no reusable component for this?

afercia: As far as I see, no. In the notifications there's a dismiss button with `<StyledIcon icon="times" ...` not reusable and with different styling.

composites/Plugin/Shared/components/YoastModal.js
afercia: Isn't it the same thing? As far as I know, React accepts all the `aria-*` attributes as props.

igorschoester:
CR 👍  ACC free 👍 
Still needs testing on premium.

igorschoester:
ACC 👍

PR #443, opened 2018-03-15
Use func proptype for icons

### LINKED ISSUE #438: HeaderIcon and Icon are defined as `string` in propType
Receiving warnings while testing https://github.com/Yoast/wordpress-seo/pull/9171.
Icon and HeaderIcon are used in the `Icon` as the `icon` property, which is a function.

They are passed as functions as the source is imported from an `SVG` to start with.

To be determined what propType should be used.
It is being fed to `Icon` which has `PropType.func` as type, thus that sounds like a logical type to choose.

Alternatively, `PropTypes.node`seems to be a more broad type that will allow for more flexible implementations.

andizer:
## Summary

* Fixes a bug where a console warning is given if an function is given as icon.

## Test instructions

* You can best test this against: https://github.com/Yoast/wordpress-seo/pull/9171

Fixes #438

### DIFF
--- a/composites/OnboardingWizard/Header.js
+++ b/composites/OnboardingWizard/Header.js
@@ -22,7 +22,7 @@ const Header = ( props ) => {
 };
 
 Header.propTypes = {
-	icon: PropTypes.string,
+	icon: PropTypes.oneOfType([ PropTypes.func, PropTypes.string ]),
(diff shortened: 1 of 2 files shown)


moorscode:
CR + Acceptance :+1:

PR #449, opened 2018-03-20
Implement collapsible panel component for classic editor sidebar

### LINKED ISSUE #436: Implement collapsible header component for classic editor sidebar
Implement a collapsible header component that can be used in the sidebar in the classic editor. In Gutenberg, we'll use the collapsible header component that's already in Gutenberg.

Please keep in mind it should work in all of the following varieties: 
34945702-7248fb0c-fa04-11e7-8b50-fbdd14a5df1e

Please reuse/refactor the code in [AnalysisCollapsible.js](https://github.com/Yoast/yoast-components/blob/develop/composites/Plugin/ContentAnalysis/components/AnalysisCollapsible.js) as much as possible.

igorschoester:
## Summary

* Adds a collapsible panel component: Collapsible and CollapsibleStateless.
* Refactored AnalysisCollapsible and AnalysisCollapsibleStateless to use the new collapsible.
* Adds a new button component: IconsButton. It supports a prefix and suffix SvgIcon besides text.
* Like the old collapsible you can vary the header level (h1, h2, h3, etc). But you always have a header now.
* Uses the IconsButton inside the header, but with different icons when open or closed.
* We do not have a WordPress sidebar widget. So this is not actually used in one.... yet.

* The stateful variant manages the `isOpen` and `Header`. The first is through the click of the button. The second is with the variable `headingLevel`.
* Another stateful difference is the children are wrapped in a styled `div` to give them a padding. Basically the AnalysisCollapsible does not want this and uses a `ul` container.

## Test instructions

* `yarn install`, `yarn start`, then go to `http://localhost:3333/` and check out the new tab: `Sidebar Collapsible`.
* Check the `Content analysis` tab too. Does this look and behave the same as before?
* Link to `wordpress-seo` and check the above in there!

Fixes #436


composites/Plugin/ContentAnalysis/components/AnalysisCollapsible.js
```diff
@@ -155,23 +132,22 @@ export class AnalysisCollapsible extends React.Component {
...
 	 */
-	isOpen() {
-		// When `isOpen` is null then the user has never opened or closed the collapsible.
-		if ( this.state.isOpen === null ) {
-			return this.props.initialIsOpen;
-		}
-
-		return this.state.isOpen === true;
+	static getHeading( props ) {
+		return wrapInHeading( StyledIconsButton, props.headingLevel );
```
maartenleenders: This function appears in this form both in 'AnalysisCollapsible' and 'Collapsible'. Not that problematic, but since it is a static function, I actually suggest calling 'wrapInHeading()' directly in the constructor and whereever else needed.

composites/Plugin/Shared/components/Collapsible.js
```diff
@@ -0,0 +1,237 @@
...
+	font-weight: 600;
+`;
+
+/**
+ * Wraps a component in a heading element with a defined heading level.
+ *
+ * @param {ReactElement} Component    The component to wrap.
+ * @param {int}          headingLevel The heading level.
+ *
+ * @returns {ReactElement} The wrapped component.
```
maartenleenders: returns a function rather than a react element

maartenleenders:
CR 👍

composites/Plugin/ContentAnalysis/components/AnalysisCollapsible.js
igorschoester: I made it static because it doesn't use any instance variables. And I made it a function because it gets used twice. What is the gain of not extraction this to a function?

composites/Plugin/Shared/components/Collapsible.js
atimmer: In this case it should be `ReactComponent`. A higher order component always receives a `ReactComponent` and returns a new `ReactComponent`. `Function` is technically correct, but a function can also be a `ReactComponent`, which it is in this case:

```jsx
const Component = ( props ) => {
    return <div>Hi!</div>;
}
```

maartenleenders:
CR 👍 I've been convinced

maartenleenders:
Acceptance 👍
I have some reservations about the way things jump around, and the blue focus square that hugs things a bit to narrowly, but none of that was introduced in this PR and it exactly copies the looks from before 👌 .

afercia:
Worth noting in Gutenberg the "arrows" have changed see https://github.com/WordPress/gutenberg/pull/5806

When closed, the arrow doesn't point to the right any longer. Gutenberg uses now "down" and "up" arrows. Will open a new issue.

PR #460, opened 2018-04-05
Implements the snippet editor in React

atimmer:
## Summary

This PR can be summarized i.n the following changelog entry:

* Add a `SnippetEditor` component. Renders the `SnippetPreview` component with editor fields to change the data.
* Add a `ReplacementVariable` component. Renders a DraftJS editor with replacement variables as entities.
* Add a `FormattedScreenReaderMessage` component Renders a message translated, but also as screen reader text. Ensures that `ScreenReaderText` can remain pure and only accept a string.

* DraftJS is used for the input fields.

	* The DraftJS mention plugin is used to auto complete replacement variables.
	* `%` is used as a trigger.
	* Replacement variables need to be passed as key value pairs. The actual shape is abstracted in a `constants.js` file.
	* We serialize the DraftJS editor state to a string on every change. Because these are small input field this is feasible. If performance suffers we need to serialize less often.
	* DraftJS is abstracted to a ReplacementVariableEditor component. This expects a string and will manage the DraftJS editor state.

* `SnippetEditor` is the main component.

	* It contains state that doesn't need to travel up the tree. Such as the hovered and active fields. And whether the editor is open.
	* The data travels upwards and is accepted as props.

* The progress bars are not measured yet. @herregroen and I thought we should solve this in YoastSEO.js. Because the assessment should be able to run independently of the snippet preview. The `SnippetEditor` component accepts the shape of an assessment result which the progress bars are rendered based upon.

* The default mode is mobile.

* We've dropped the `SnippetPreview` example because the `SnippetEditor` example covers that completely.

* Added 3 colors to `colors.scss` and `colors.json` that are used for the carets.

### Also includes some quality of life changes

* Change devtool to `cheap-module-eval-source-map` which make sourcemaps work.
* Add NameModulesPlugin so you can see which modules have been hot reloaded.
* Use `.babelrc` in the webpack config.
* Add `CaseSensitivePathsPlugin`, this makes us aware of case sensitivity issues earlier.
* Adds new pattern for exposing public components. Every separate part can expose its components using an `index.js`, these will be aggregated in the root `index.js`. I haven't refactored all other components.
* Setup test coverage for all files. This means now files that have 0% coverage are also included in the coverage. This drastically reduced the coverage, but it does mean that it is more accurate.

## Test instructions

* Turn on the local example using `yarn start`.
* Go to the `Snippet preview` menu item.
* Test that the snippet editor works in the same way as the old one (currently in the plugin).


App.js
```diff
@@ -13,6 +12,7 @@ import SidebarCollapsibleWrapper from "./app/SidebarCollapsibleWrapper";
 // Required to make Material UI work with touch screens.
 import injectTapEventPlugin from "react-tap-event-plugin";
 import Checkbox from "./composites/Plugin/Shared/components/Checkbox";
+import SnippetEditor from "./app/SnippetEditorWrapper";
```
moorscode: `Wrapper` naming does not seem to be explanatory, have you considered using `Example`?

composites/Plugin/SnippetEditor/components/ReplacementVariableEditor.js
```diff
@@ -0,0 +1,167 @@
...
+import { replacementVariablesShape } from "../constants";
+
+/**
+ * Creates a DraftJS editor state from a string.
+ *
+ * @param {string} content The content to turn into editor state.
+ *
+ * @returns {EditorState} The editor state.
+ */
+const createEditorState = flow( [
```
moorscode: `flow` is a magical thing, I cannot really guess what is happening here. Is there a way to make this more usable/readable?

composites/Plugin/SnippetEditor/components/ReplacementVariableEditor.js
```diff
@@ -0,0 +1,167 @@
...
+ * @param {EditorState} The current editor state.
+ *
+ * @returns {string} The serialized editor state.
+ */
+const serializeEditorState = flow( [
+	convertToRaw,
+	serializeEditor,
+] );
+
+class ReplaceVarEditor extends React.Component {
```
herregroen: This class is missing documentation.

composites/Plugin/SnippetEditor/components/ReplacementVariableEditor.js
```diff
@@ -0,0 +1,167 @@
...
+	/**
+	 * Constructs the replacement variable editor for use.
+	 *
+	 * @param {Object} props The props to instantiate this editor with.
```
herregroen: It would be nice to fully specify all the keys you are able to pass to props.

composites/Plugin/SnippetEditor/components/ReplacementVariableEditor.js
```diff
@@ -0,0 +1,167 @@
...
+					onAddMention={ this.onAddMention }
+				/>
+			</React.Fragment>
+		);
+	}
+}
+
+ReplaceVarEditor.propTypes = {
+	content: PropTypes.string.isRequired,
+	onChange: PropTypes.func,
```
moorscode: Consider placing this close to the other event-props.

moorscode: This is not required, but also not supplied in the defaultProps.

composites/Plugin/SnippetEditor/components/ReplacementVariableEditor.js
```diff
@@ -0,0 +1,167 @@
...
+		super( props );
+
+		const { content } = this.props;
+
+		this.state = {
+			editorState: createEditorState( content ),
+			replacementVariables: props.replacementVariables,
+		};
+
+		this._serializedContent = content;
```
moorscode: The sudden appearance of `serialized` is not evident from the code, could you elaborate on this in a comment or maybe choose different variable names.

moorscode: Suggestions: use `raw` as definition, as this is unprepared/touched.

composites/Plugin/SnippetEditor/components/ReplacementVariableEditor.js
```diff
@@ -0,0 +1,167 @@
...
+
+		this.onChange = this.onChange.bind( this );
+		this.onSearchChange = this.onSearchChange.bind( this );
+
+		this.mentionsPlugin = createMentionPlugin( {
```
herregroen: Given that the mentions plugin is used for replacement variables it would probably be helpful to add a comment here why we're using the mentions plugin.

composites/Plugin/SnippetEditor/components/SnippetEditor.js
```diff
@@ -0,0 +1,387 @@
...
+/* Internal dependencies */
+import SnippetPreview, { MODE_DESKTOP, MODE_MOBILE, MODES } from "../../SnippetPreview/components/SnippetPreview";
+import SnippetEditorFields from "./SnippetEditorFields";
+import { Button } from "../../Shared/components/Button";
+import SvgIcon from "../../Shared/components/SvgIcon";
+import colors from "../../../../style-guide/colors.json";
+import { lengthAssessmentShape, replacementVariablesShape } from "../constants";
+import FormattedScreenReaderMessage from "../../../../a11y/FormattedScreenReaderMessage";
+
+const SwitcherButton = Button.extend`
```
herregroen: This component should be documented and probably other styled components should be documented.

composites/Plugin/SnippetEditor/components/ReplacementVariableEditor.js
atimmer: After you know what `flow` does, this code is more readable than its alternatives.

composites/Plugin/SnippetEditor/components/ReplacementVariableEditor.js
```diff
@@ -0,0 +1,195 @@
...
+/**
+ * A replacement variable editor. It allows replacements variables as tokens in
+ * its editor. It's a small shell on top of DraftJS.
+ */
+class ReplacementVariableEditor extends React.Component {
+	/**
+	 * Constructs the replacement variable editor for use.
+	 *
+	 * @param {Object} props                        The props to instantiate this
+	 *                                              editor with.
```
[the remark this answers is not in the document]
atimmer: The [WordPress JavaScript documentation standards](https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/javascript/#line-wrapping) state that lines should be wrapped at 80 characters. Unless they are indented.

composites/Plugin/SnippetEditor/components/SnippetEditor.js
```diff
@@ -0,0 +1,334 @@
...
+
+	/**
+	 * Focuses the preview on the given field.
+	 *
+	 * @param {String} field the name of the field to focus
+	 *
+	 * @returns {void}
+	 */
+	setFieldFocus( field ) {
+		field = this.mapFieldToEditor( field );
```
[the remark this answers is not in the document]
atimmer: Because that is easier to read.

igorschoester:
CR 👍

afercia:
Visually, hovering/focusing the mobile icon makes it "jump".

afercia:
Important: a far as I know, the *slug* field isn't meant to accept template variables, correct? IN this case, ti shouldn't be announced as combobox and shouldn't have any of the HTMl attributes related to autocomplete, activedescendant. etc. It should be just an ordinary input field, and exposed as such to assistive technologies.


Positioning issues:
in some edge cases, the list of suggestions might be displayed off screen, for example when the caret in the "textarea" is close to the right edge. Not sure it the mentions plugin has anything to handle the positioning:

screen shot 2018-04-23 at 16 02 10

atimmer:
Merging this. All the issues uncovered by Andrea I will create as issues to be fixed before release of the plugin.

afercia:
I forgot to mention: there are no instructions on how to use the suggestions, e.g. type `%` and see what happens. I'd say even sighted user would need to be instructed. I'd suggest to try the simplest thing: add a paragraph below the fields `type the percentage character % for a list of suggested template variables` (or something along those lines) and target the paragraph with `aria-describedby` on the fields.

PR #465, opened 2018-04-12
Collapsible panel using up/down

### LINKED ISSUE #463: Collapsible panel should use down / up arrows
Gutenberg used to use right / down arrows fir the collapsible panels so #449 implemented arrows that way. However arrows have changed in Gutenberg see https://github.com/WordPress/gutenberg/pull/5806 and they now are down / up. This is to be consistent with similar panels in WordPress.

I guess the collapsible panel should update too and use down / up arrows.

boblinthorst:
## Summary

* The collapsible panel now uses up/down arrows instead of the right/down arrows it used to use. Also made sure the direction of the arrows matches gutenberg.

I tried quickly scaling down the size of the arrows, but couldn't find out quickly how to do it best. If anyone has any suggestions how to scale down svg images, please tell.

* The gutenberg dashicons won't be used in this pull yet, because they're not available outside gutenberg yet.

## Test instructions

* Check if the arrows in this branch match the arrows used on the post page by gutenberg.

Fixes #463

### DIFF
--- a/composites/Plugin/Shared/components/Collapsible.js
+++ b/composites/Plugin/Shared/components/Collapsible.js
@@ -237,7 +237,7 @@ Collapsible.defaultProps = {
 	initialIsOpen: false,
 	prefixIconColor: colors.$black,
 	suffixIcon: "arrow-down",
-	suffixIconCollapsed: "arrow-right",
+	suffixIconCollapsed: "arrow-up",
 	suffixIconColor: colors.$black,
 };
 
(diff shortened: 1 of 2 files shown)


igorschoester:
CR & ACC 👍 
To reiterate on the size / dashicons part of the issue. We will be awaiting [what Wordpress does with the dashicons](https://github.com/WordPress/dashicons/issues/255).

PR #473, opened 2018-04-18
Add react-axe to app.js

IreneStr:
## Summary

* Added react-axe to app.js to enable a11y testing.


IreneStr:
Closed, because we're not going to implement this (for now). Axe's suggestions do not always match WCAG guidelines, and the examples in app.js are not always 100% the same as what components will be used in the plugin.

Issue #485, opened 2018-05-02
Breadcrumbs in React snippet preview are cut off by question mark and hash


IreneStr:
### How can we reproduce this behavior?
1. Use define( 'YOAST_FEATURE_SNIPPET_PREVIEW', true ); to show the React SnippetPreview.
2. Open the snippet editor of the React snippet preview and make sure it's showing the mobile view. Write a slug and look at the breadcrumbs.
3. Add a question mark somewhere in the middle of your slug.
4. See that the breadcrumb only contains the part before the question mark.

The same happens when you use a #.

### To do
Because WordPress strips ? and # from the slug in the permalink editor, we should do that as well.

afercia:
Seems to me the issue here is because `getBreadcrumbs()` uses `parse` from the `node-url` package which considers anything that comes after a `?` a query string and anything that comes after a `#` an URL fragment identifier.  It correctly returns them in the `search` and `hash` parts of the parsed URL but we're not using them.
So yes, I guess the most reasonable option is to just strip out any `?` and `#` from the URL _before_ passing it to `parse( url )`.

PR #495, opened 2018-05-04
Strip out question mark and hash characters from the snippet preview …

afercia:
Strips out any `?` and `#` from the snippet preview URL before passing it to node-url `parse()`.

To test:
- in the standalone example or yarn-link to wpseo:
- if testing on WordPress, make sure your `wp-config.php` sets the constant `define( 'YOAST_FEATURE_SNIPPET_PREVIEW', true );`
- if testing on WordPress, build the wpseo plugin js
- edit the snippet preview
- in the slug field, enter a slug that contains `?` and `#` characters
- check the slug in the snippet preview is rendered correctly, without `?` and `#` and with no missing parts

Fixes https://github.com/Yoast/yoast-components/issues/485


Xyfi:
CR :+1:

Xyfi:
Acceptance :+1:

PR #509, opened 2018-05-07
Make mobile arrow visible 

### LINKED ISSUE #504: Snippet preview: mobile description "arrow" is not visible
Compare with the desktop snippet preview: hover or click the description and the "arrow" appears on the left:

screen shot 2018-05-07 at 12 56 44

Instead, in the mobile snippet preview, the arrow is not visible:

screen shot 2018-05-07 at 12 56 53

See `overflow: hidden` used on the `MobileDescription`, which prevents the arrow to be visible.

Xyfi:
by removing `overflow:hidden;` when description has been shortened to the right length.

## Summary

* N/A

* When the description has been shortened to teh right length, `overflow:hidden;` is removed from the mobile description `div`. Since it can take a short while for the description to be cut down to the right length the arrow will not show for a split second when the description is updated.

## Test instructions

* Open the test application.
* Check that the arrow is visible when clicking on the mobile description field.

Fixes #504

### DIFF
--- a/composites/Plugin/SnippetPreview/components/SnippetPreview.js
+++ b/composites/Plugin/SnippetPreview/components/SnippetPreview.js
@@ -133,7 +133,8 @@ export const DesktopDescription = styled.div.attrs( {
 const MobileDescription = styled( DesktopDescription )`
 	line-height: 1em;
 	max-height: 4em;
-	overflow: hidden;
+	/* Hide overflow as long as description length is being calculated */
+	overflow: ${ props => props.overflowHidden ? null : "hidden" };
 `;
 
 const MobilePartContainer = styled.div`
@@ -243,6 +244,7 @@ export default class SnippetPreview extends PureComponent {
 		this.state = {
 			title: props.title,
 			description: props.description,
+			descriptionFits: false,
 		};
 
 		this.setTitleRef       = this.setTitleRef.bind( this );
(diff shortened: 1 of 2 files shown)


composites/Plugin/SnippetPreview/components/SnippetPreview.js
```diff
@@ -133,7 +133,8 @@ export const DesktopDescription = styled.div.attrs( {
 const MobileDescription = styled( DesktopDescription )`
 	line-height: 1em;
 	max-height: 4em;
-	overflow: hidden;
+	/* Hide overflow as long as description length is being calculated */
+	overflow: ${ props => props.overflowHidden ? null : "hidden" };
```
afercia: For better clarity, I'd expect the other way around: if there's a prop overflowHidden then set to "hidden".

composites/Plugin/SnippetPreview/components/SnippetPreview.js
```diff
@@ -605,7 +613,8 @@ export default class SnippetPreview extends PureComponent {
 						             onClick={ onClick.bind( null, "description" ) }
 						             onMouseOver={ partial( onMouseOver, "description" ) }
 						             onMouseLeave={ partial( onMouseLeave, "description" ) }
-						             innerRef={ this.setDescriptionRef } >
+						             innerRef={ this.setDescriptionRef }
+						             overflowHidden={ this.state.descriptionFits }>
```
afercia: See above, I'd ex[ect the other way around: overflowHidden true when description does _not_ fit.

Issue #516, opened 2018-05-09
Snippet preview: highlightKeyword doesn't work (missing keyword prop)


afercia:
Seems to me the keyword is not highlighted in the url and description (if present) for a couple reasons:

- `SnippetPreview` expects a `keyword` prop, seems to me it's not passed anywhere and the default is empty string
- even if it's passed, the highlighting would work only in the desktop view (edit: won't work on the url, while it works on the description). Not sure if that's intentional, see https://github.com/Yoast/yoast-components/blob/c52c1242e5b79238a86ea2cab5c1fa5ae314c014/composites/Plugin/SnippetPreview/components/SnippetPreview.js#L475

atimmer:
The second point I don't know about. Should we highlight the keyboard in breadcrumb style?

afercia:
Honestly I'm not sure. Does Google still highlights search terms in the url in the first place? /Cc @jdevalk

boblinthorst:
Though not a be all end all, a quick tests shows that google doesn't seem to highlight the search term in the url. Here's a google search for the term "awesome"
<img width="580" alt="awesome_-_google_search" src="https://user-images.githubusercontent.com/11849359/39923700-0fbef8fc-5525-11e8-8f82-60aaa82fffda.png">

boblinthorst:
When searching with google for the term `awesome`, this is what a result looks like:

<img width="580" alt="awesome_-_google_search" src="https://user-images.githubusercontent.com/11849359/39926999-33c3baba-5531-11e8-97e7-feb7407872a6.png">

the search term seems to be highlighted in the description, not in the title or the slug. 

When doing the same search on a mobile phone, the keyword isn't highlighted anywhere.

PR #519, opened 2018-05-09
Add replace diacritics to Snippet Preview and preserve transliteration.

afercia:
## Summary

* Replace diacritics in the Snippet Preview url.

- adding `replaceDiacritics` from YoastSEO.js is trivial, making it coexist with the transliteration and "highlightKeyword" features is a bit more complicated; I may have missed something about the intended functionality please double check /Cc @IreneStr 
- this PR tries to determine when a text that has been cleaned up from diacritics is then passed to `highlightKeyword()`
- if that is the case, `highlightKeyword` and will match against a cleaned up keyword as well
- the transliteration should be safe and still runs on the non-cleaned up keyword and text

## Test instructions

- testing is a bit weird because right now the SnippetPreview doesn't get passed a `keyword` and `locale` props which are necessary to make it work propery
- suggestion: run `yarn start` and check the standalone snippet preview on localhost:3333
- in `composites/Plugin/SnippetPreview/components/SnippetPreview.js` manually change the default props `keyword` and `locale` to `kåyword` and `da`
- switch do the Desktop preview (highlighting in the Mobile preview doesn't work, see #516)
- edit the Snippet Preview: in the Slug and Meta description enter some text containing both the original keyword `kåyword` and the transliterated one `kaayword`
- verify both are highlighted in the preview url and description
- of course, the url will show the "cleaned up" `kayword` instead of `kåyword`
- change locale and keyword, and play with other diacritics, see attached [diacritics.txt](https://github.com/Yoast/yoast-components/files/1987809/diacritics.txt)

Screenshot:

screen shot 2018-05-09 at 14 07 32

Fixes https://github.com/Yoast/wordpress-seo/issues/9674

### DIFF
--- a/composites/Plugin/SnippetPreview/components/SnippetPreview.js
+++ b/composites/Plugin/SnippetPreview/components/SnippetPreview.js
@@ -4,6 +4,7 @@ import styled from "styled-components";
 import interpolateComponents from "interpolate-components";
 import transliterate from "yoastseo/js/stringProcessing/transliterate";
 import createWordRegex from "yoastseo/js/stringProcessing/createWordRegex";
+import replaceSpecialCharactersAndDiacritics from "yoastseo/js/stringProcessing/replaceDiacritics";
 import PropTypes from "prop-types";
 import truncate from "lodash/truncate";
 import partial from "lodash/partial";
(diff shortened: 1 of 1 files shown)


danielFangstrom:
CR 👍

PR #521, opened 2018-05-11
Pass locale prop from snippet editor to snippet preview

### LINKED ISSUE #518: Snippet preview: missing "locale" prop
As far as I see, the SnippetPreview component needs a `locale` prop to work **prop**erly. For example, see `highlightKeyword > transliterate` where the transliteration needs a locale otherwise it won't work.

Seems to me there's just a default locale prop `locale: "en_US"` and a prop is not passed anywhere. I guess it should be passed by the parent component `SnippetEditor`.

Worth also noting the transliteration needs a two characters locale, e.g. `da` for Danish, while it seems we're using an ISO/Posix `language_region` notation e.g. `"en_US"`.

Edit: additionally, I guess the actual locale to use should be related to the site language, not the user language.

Xyfi:
## Summary
Makes the `SnippetEditor` accept a locale prop that is passed down to the `SnippetPreview`.
Also changed the default locale from `en_US` to `en`.

* Not applicable

* 

## Test instructions

* 

Fixes #518

### DIFF
--- a/app/SnippetEditorExample.js
+++ b/app/SnippetEditorExample.js
@@ -47,7 +47,7 @@ export default class SnippetEditorExample extends Component {
 			" Title Snippet Title Snippet Title Snippet Title Snippet Title",
 			url: "https://local.wordpress.test/welcome-to-the-gutenberg-editor-2/",
 			slug: "welcome-to-the-gutenberg-editor-2",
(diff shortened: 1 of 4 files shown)


danielFangstrom:
CR 👍

danielFangstrom:
ACC 👍

PR #522, opened 2018-05-11
Keyword highlighting in the description

### LINKED ISSUE #516: Snippet preview: highlightKeyword doesn't work (missing keyword prop)

boblinthorst:
Pass along the keyword prop to the snippet preview to allow for keyword highlighting. Also added keyword as the keyword prop for the standalone version, to have it function there.

## Summary

*N/A

* Passed it along as a separate prop, could have also made it part of the data shape.

## Test instructions

* On the standalone, add the word 'keyword' to the description, and check if it's highlighted.

Fixes #516

### DIFF
--- a/app/SnippetEditorExample.js
+++ b/app/SnippetEditorExample.js
@@ -199,6 +199,7 @@ export default class SnippetEditorExample extends Component {
 				replacementVariables={ replacementVariables }
 				titleLengthAssessment={ titleLengthAssessment }
 				descriptionLengthAssessment={ descriptionLengthAssessment }
+				keyword="keyword"
 			/>
 
 			<h2>Test Sliders</h2>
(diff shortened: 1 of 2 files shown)


danielFangstrom:
CR 👍

danielFangstrom:
ACC 👍

PR #526, opened 2018-05-14
Stories/9590 add trailing slashes to react snippet editor

danielFangstrom:
## Summary

* Adds trailing slashes to the url in the desktop mode

* Checks the slug for trailing slashes and adds one in case it is missing. Fixes the bug in the old snippet editor that added additional slashes when already present.

## Test instructions

* Run `yarn start` and go the stand-alone snippet preview on localhost:3333
* Edit snippet and type something in the slug field
* There should be a trailing slash in the url, even if you did not end the slug text with one
* There should be no extra slashes if you have one or more slashes at the end of the slug text (i.e. if you have two slashes in the slug text, two should be present in the url).

Fixes https://github.com/Yoast/wordpress-seo/issues/9590

### DIFF
--- a/composites/Plugin/SnippetPreview/components/SnippetPreview.js
+++ b/composites/Plugin/SnippetPreview/components/SnippetPreview.js
@@ -253,6 +253,16 @@ function highlightKeyword( locale, keyword, text, cleanText ) {
 	} );
 }
 
+/**
+ * Returns if a url has a trailing slash or not.
+ *
+ * @param {string} url The url to check for a trailing slash.
+ * @returns {boolean} Whether or not the url contains a trailing slash.
+ */
+function hasTrailingSlash( url ) {
+	return url.indexOf( "/" ) === ( url.length - 1 );
+}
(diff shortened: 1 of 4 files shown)


composites/Plugin/SnippetPreview/components/SnippetPreview.js
```diff
@@ -516,11 +527,19 @@ export default class SnippetPreview extends PureComponent {
 		if ( this.props.mode === MODE_MOBILE ) {
 			urlContent = this.getBreadcrumbs( cleanUrl );
 		} else {
-			urlContent = highlightKeyword( locale, keyword, url, cleanUrl );
+			/*
```
boblinthorst: comment block start is off by 1 space

boblinthorst: My bad, it was the other way around. All the other comment stars + the end should be indented with 1 space, the first line was fine.

PR #535, opened 2018-05-17
Better encode the SVG carets in the snippet preview and editor

### LINKED ISSUE #532: Selection carets not working in firefox
### Please give us a description of what happened.

The selection carets are not working in firefox. In the screenshot, I have selected the meta description, but the visual aid to make that clear is missing:

yoast_components_and_my_short_story___ _local_wordpress_test_ _wordpress

### Please describe what you expected to happen and why.

For them to be present, as they are in chrome, for instance. Here I'm selecting the title in chrome:

yoast_components

### How can we reproduce this behavior?
See above headers.

### Technical info
* yoast-components version: develop

afercia:
## Summary

* Better encode SVG images used as CSS background.

- replaces `encodeURI` with `encodeURIComponent`
- not 100% sure this is the best method but `encodeURI` does not encode, for example, `:`, `/`, `=`, and they need to be encoded to make Firefox and IE 11 happy
- `encodeURIComponent` encodes them, even though it still doesn't encode, for example, dots `.` and hyphens `-`
- regardless, it seems to work now

## Test instructions

- on the standalone version, run `yarn start` and go to http://localhost:3333/
- please test with **all** browsers the carets are shown properly

I've tested it on:
mac Chrome, Firefox, Safari, Opera
windows IE 11, Edge, Chrome, Firefox

Note:
at some point I was getting an error on Edge, see screenshot below, but then it disappeared and I wasn't able to reproduce it again:

screenshot 131

Fixes #532

### DIFF
--- a/composites/Plugin/SnippetEditor/components/SnippetEditorFields.js
+++ b/composites/Plugin/SnippetEditor/components/SnippetEditorFields.js
@@ -13,7 +13,7 @@ import ControlledInput from "./ControlledInput";
 
 import colors from "../../../../style-guide/colors";
 
-const angleRight = ( color ) => "data:image/svg+xml;charset=utf8," + encodeURI(
+const angleRight = ( color ) => "data:image/svg+xml;charset=utf8," + encodeURIComponent(
 	'<svg width="1792" height="1792" viewBox="0 0 1792 1792" xmlns="http://www.w3.org/2000/svg">' +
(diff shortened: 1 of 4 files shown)


danielFangstrom:
CR 👍 
Not encoding dots and hyphens should not cause any issues as far as I can tell

danielFangstrom:
ACC 👍 
Tested all the browsers mentioned in the description and on windows as well, and the carets looked correct and consistent in all cases. Could not reproduce the error in edge.

PR #537, opened 2018-05-17
Replace replace vars before measuring the title and description length

IreneStr:
## Summary

* Replace the replace vars before measuring the title and description length to get the correct length.

## Technical choices
- The replaceVariables function from the global is declared twice. When I put it on top of the file, it (obviously) couldn't be found, but the constructor was not a good location either. Suggestions are welcome, but please don't spend too much time.


boblinthorst:
Acceptance :+1:

PR #538, opened 2018-05-17
Set `currentColor` as default icon color

### LINKED ISSUE #445: SvgIcon: consider to use `currentColor` as default icon color
The SvgIcon component doesn't have a default value for the icon color. This way, all icons will have a default black color, quick test:

screen shot 2018-03-16 at 11 26 09

This requires to pass every time a color prop. In this cases it would be nice to have the icon of the same color of the text by default.

I'd propose to consider a smarter default color prop, à la Gutenberg. Using a default prop value `currentColor` will result in a CSS property `fill: currentColor;` which inherits the color of the surrounding text. 

See
https://www.w3.org/TR/css-color-3/#currentcolor
https://caniuse.com/#search=currentcolor

needs #719

Rahmon:
## Summary

* Add `currentColor` as default icon color

* Set default value for color prop

## Test instructions

* This has been tested manually on Chrome.

Fixes #445


afercia:
The build is failing because the snapshots need to be updated. This can be done running `yarn test -u` and committing the updated snapshots.

Also, it would be nice to address #719 first to be able to effectively test al the icons.

Rahmon:
Ok. I'll update the PR.

IreneStr:
@Rahmon Just a quick reminder to update your branch 😄. If you update it before next Friday, we'll do our best to include your patch in our 8.4 release.

Rahmon:
@IreneStr 
I've updated the PR.

Xyfi:
CR :+1:

Xyfi:
Acceptance :+1:

PR #539, opened 2018-05-17
Fix snippet editor JS error with invalid RawDraftContentState

### LINKED ISSUE #534: Snippet editor: JS Uncaught Error: invalid RawDraftContentState
To reproduce:
- test the standalone version (yarn start and go to localhost:3333)
- open the snippet editor
- click on the title or the description field 
- regardless of where the caret is placed in the field...
- press Backspace
- the whole page goes blank: `Uncaught Error: invalid RawDraftContentState`

igorschoester:
## Summary

* Fixes a problem with the Snippet Editor where it would crash when receiving new content.

* Unserialize the new content in componentWillReceiveProps.
* Add extra check to test equality for when we received the old content as if it was new.
* Add ReplacementVariables to the componentWillReceiveProps as well to work correctly with the changing custom fields replacement variables.

## Test instructions

* In the standalone app, it should no longer crash immediately after changing the title.
* In combination with the Yoast SEO plugin:
  - Test if you no longer see the JS Error (did not occur always like in the standalone).
  - When you add, update or remove a custom field in a post you should see this reflected in the 
 replacement variables in the Snippet Editor right away. Where removing the custom field if you have it in the title you should see the `%%cf_<name_of_custom_field>%%` instead.

Fixes #534

### DIFF
--- a/composites/Plugin/SnippetEditor/components/ReplacementVariableEditor.js
+++ b/composites/Plugin/SnippetEditor/components/ReplacementVariableEditor.js
@@ -215,7 +215,7 @@ class ReplacementVariableEditor extends React.Component {
 
 		if (
 			( nextProps.content !== this._serializedContent && nextProps.content !== content ) ||
-		    nextProps.replacementVariables !== replacementVariables
+			nextProps.replacementVariables !== replacementVariables
 		) {
 			this._serializedContent = nextProps.content;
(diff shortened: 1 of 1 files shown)


maartenleenders:
CR 👍

CarolineGeven:
acceptance done

PR #540, opened 2018-05-17
Fix snippet preview overflow with long text.

afercia:
## Summary

* Improve the Snippet Preview rendering with long text.

- unless I'm missing something, the description should have maximum 4 lines of text
- adjusts the max-height to a proper value, based on the line height
- also improves some indentation, as we shouldn't have lines with a mix of tab characters and spaces

## Test instructions

- to test on the standalone version, I've merged develop so the fix from #539 is in
- run `yarn start` and go on localhost:3333
- first, observe the behavior in the title field
- depending on the text added, you may see the overflowing word appear for some milliseconds and then disappear (suggestion: try repeatedly with very short words)
- then compare with the description field behavior
- the max number of lines is 4: when adding overflowing words, they disappear almost instantly (behavior should be the same of the title field)
- optionally, yarn link to the plugin and test in the meta box

Fixes https://github.com/Yoast/wordpress-seo/issues/9735

### DIFF
--- a/composites/Plugin/SnippetEditor/components/SnippetEditorFields.js
+++ b/composites/Plugin/SnippetEditor/components/SnippetEditorFields.js
@@ -60,7 +60,7 @@ const InputContainer = styled.div.attrs( {
 	font-family: Arial, Roboto-Regular, HelveticaNeue, sans-serif;
 	font-size: 14px;
 	margin-top: 5px;
-	
+
 	&::before {
 		display: block;
 		position: absolute;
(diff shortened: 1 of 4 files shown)


boblinthorst:
CR: ok 👍 I did a quick check with the previous snippet editor. It doesn't seem to have a limit of how many lines are allowed to be in the snippet editor, so we're changing the behaviour a bit, but based on Anton's commit message, it's intended (And you really need to try to make your meta description exceed four lines )

boblinthorst:
Acceptance: ok 👍 They indeed disappear quicker in the description now compared to the title.

PR #542, opened 2018-05-18
Snippet editor improve fields focus and blur

### LINKED ISSUE #503: Snippet editor: fields shouldn't look active when not focused
In the new React snippet editor:
- click on a field
- then click outside of the field or on any other focusable element that is not a snippet editor field
- the field still looks "focused" while actually it isn't:

screen shot 2018-05-07 at 10 47 49

afercia:
## Summary

* Avoid the snippet editor fields to stay in a `isActive` state when blurred

- adds an `onBlur` handler to set the `activeField` to `null` when the fields are blurred
- this required some adjustments to the way clicking on the preview elements works, it now uses `onMouseUp` instead of `onClick`, please refer to inline comment (we might want to re-think how functions are passed and bound to the snippet preview elements)
- changes `onMouseOver` to `onMouseEnter` as the former happens too often
- always focuses the fields when clicking on their labels: this was prevented by the condition `activeField !== prevActiveField`
- tries to reduce the number of times the fields get focused: this was happening under the hood also when hovering the snippet preview (and potentially it happens for any props update)
- removes the usage of lodash `partial()` for consistency and readability

## Test instructions

- run the standalone yoast-components (`yarn start` and then go to localhost:3333)
- click on an element in the preview
- verify the corresponding field in the editor gets focused
- verify that when clicking outside the preview or outside the fields, the fields don't stay in an "active" state any longer
- same when tabbing away from the fields
- verify that clicking on the preview elements always focuses the associated field correctly

Note 1: I see an issue on Safari, but it happens also on `develop`, will open a separate issue.

Note 2: there's a problem with the replaced variables within the fields in IE 11, will open a separate issue

Fixes #503


afercia:
Rebased.

afercia:
OK I think there's a change to implement about events, re-renderings, and focusing the fields that needs to be made for this to work.

This PR adds an `onBlur` handler, so that when clicking outside the preview or outside one of the editor fields, the active fields blurs and is no "active" any longer. Same happens when tabbing away from a field. This is good.

However, this initially broke the `click` behavior on the preview elements. Focusing happens on `componentDidUpdate()` so what the component currently does is focusing again whenever any prop updates. Right now, it focuses again also when hovering 😬you just can't see it, but `focusOnActiveFieldChange()` runs also when hovering and, potentially, when any other future prop updates. That's a bit wild.

It's a matter of timings and re-renderings, as now the snippet preview re-renders also when the `activeField` prop updates `onBlur`. Basically, after an `onBlur`, the click event didn't happen at all, I guess because a new function is created and bound when re-rendering (?).

Using `onMouseUp` instead of `onClick` seems to fix it, will test a bit more to double check.

Worth noting the behavior on click will match the one in the classic snippet editor: when clicking for the first time on a snippet preview element, the related field gets focused:

clicked

when clicking again, and _while the mouse is still pressed_, the actual focused element is the preview element, so in that exact moment there's no focused field:

while clicking

Only when the mouse is released, the field gets focused again.

afercia:
Forgot to mention this PR removes a bunch of things from `SnippetEditorExample.js` as they seem to me not used.

composites/Plugin/SnippetEditor/components/SnippetEditorFields.js
```diff
@@ -162,37 +162,30 @@ class SnippetEditorFields extends React.Component {
...
-	 * @param {Object} prevProps The previously received props.
+	 * For example, the component will update when clicking on the field labels.
+	 * In this case, we need to focus again the field.
 	 *
+	 * @param {Object} prevProps The previous props.
 	 * @returns {void}
 	 */
 	componentDidUpdate( prevProps ) {
-		this.focusOnActiveFieldChange( prevProps.activeField );
+		if ( prevProps.activeField === this.props.activeField ) {
```
Xyfi: This should ideally be
```
if ( prevProps.activeField !== this.props.activeField ) {
  this.focusOnActiveFieldChange();
}
```

Xyfi:
I also tested it and it worked fine.

afercia:
Last commit removes a test and condition for `onMouseLeave()` as when mouse-leaving an element there's no way to mouse-leave another one before actually mouse-enter it.

afercia:
@Xyfi thanks, improved that conditional.

igorschoester:
CR 👍

afercia:
As @igorschoester correctly pointed out, the `onMouseUp` won't work on mobile devices.

afercia:
Correction (and for reference): It seems, though it there’s no official standard, the de-facto standard is mobile devices fires also mouse events. See 
https://developer.mozilla.org/en-US/docs/Web/API/Touch_events/Supporting_both_TouchEvent_and_MouseEvent
https://w3c.github.io/touch-events/#mouse-events

igorschoester:
Acceptance 👍

PR #543, opened 2018-05-22
9771/slug from prop instead of from state

boblinthorst:
## Summary

* Changed slug to prop slug.

* Removed controlled input. Changed it to styled component.

## Test instructions

* see sibling pr. see parent

Fixes # https://github.com/Yoast/wordpress-seo/issues/9771


igorschoester:
CR & Acceptance 👍

PR #544, opened 2018-05-22
Implement displaydata support

maartenleenders:
## Summary

* Use the displayData field from the redux store to inform what text should be displayed in the snippet preview.

* See https://github.com/Yoast/wordpress-seo/pull/9806

## Test instructions

* See https://github.com/Yoast/wordpress-seo/pull/9806

Fixes https://github.com/Yoast/wordpress-seo/issues/9752

### DIFF
--- a/composites/Plugin/SnippetEditor/components/SnippetEditor.js
+++ b/composites/Plugin/SnippetEditor/components/SnippetEditor.js
@@ -325,13 +325,13 @@ class SnippetEditor extends React.Component {
 	 *
 	 * @returns {Object} The data for the preview.
 	 */
-	mapDataToPreview( originalData ) {
+	mapDataToPreview( originalData, displayData ) {
 		const { baseUrl, mapDataToPreview } = this.props;
 
 		const mappedData = {
 			title: this.processReplacementVariables( originalData.title ),
 			url: baseUrl.replace( "https://", "" ) + originalData.slug,
-			description: this.processReplacementVariables( originalData.description ),
+			description: this.processReplacementVariables( originalData.description || displayData.description ),
 		};
 
 		if ( mapDataToPreview ) {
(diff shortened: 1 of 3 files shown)


composites/Plugin/SnippetEditor/components/SnippetEditor.js
```diff
@@ -322,16 +322,17 @@ class SnippetEditor extends React.Component {
 	 * Maps the data from to be suitable for the preview.
 	 *
 	 * @param {Object} originalData The data from the form.
+	 * @param {Object} displayData  The data from the redux store.
```
afercia: I'd consider to explain a bit better what the `data` from redux store are.

afercia:
CR 🚧 

A few minor coding standards

composites/Plugin/SnippetPreview/components/SnippetPreview.js
```diff
@@ -632,6 +645,7 @@ export default class SnippetPreview extends PureComponent {
 					{ ...outerContainerProps }
 				>
 					<MobileDescriptionOverflowContainer
+						isDescriptionGenerated={ this.state.isDescriptionGenerated }
```
boblinthorst: This is only called when mode is mobile, will it also be called when it is in desktop mode?

boblinthorst:
CR: made a single comment. not sure if correct.

composites/Plugin/SnippetPreview/components/SnippetPreview.js
maartenleenders: Yes, it is located in the `outerContainerProps` object. I don't know why the mobile container is wrapped twice, but I think both the `MobileDescriptionOverflowContainer` and the `MobileDescriptionWithCaret` need to have the `isDescriptionGenerated` prop to set the right colors.

PR #545, opened 2018-05-22
Full meta description editor clickable

Xyfi:
## Summary

* All input wrappers in the `SnippetEditorFields` feel and behave like an input and move focus to the wrapped input.

*

## Test instructions

* Click the lower part of the meta description field and see whether it focusses the editor.

Fixes Yoast/wordpress-seo#9798


IreneStr:
Acceptance 👍

composites/Plugin/SnippetEditor/components/SnippetEditorFields.js
```diff
@@ -205,36 +205,45 @@ class SnippetEditorFields extends React.Component {
...
+						isHovered={ hoveredField === "title" }>
 						<ReplacementVariableEditor
 							content={ title }
 							onChange={ content => onChange( "title", content ) }
 							onFocus={ () => onFocus( "title" ) }
 							replacementVariables={ replacementVariables }
 							ref={ ( ref ) => this.setRef( "title", ref ) }
-							ariaLabelledBy={ this.uniqueId + "-title" }
+							ariaLabelledBy={ titleLabelId }
 						/>
```
afercia: Not sure if there's a specific reason for this change, it will result in a value that's just a number pointing to the ID above at line 230 that's just a number, see screenshot below:

screen shot 2018-06-18 at 10 21 08

I guess it would be safer to have a more qualified ID/aria-labelledby value to avoid potential conflicts with other IDs (you never know). This should apply also to the description field. As far as I see, the slug field still uses a more qualified value `this.uniqueId + "-slug" `. Will open a separate issue.

PR #546, opened 2018-05-23
Move translation to React render method

atimmer:
## Summary

* Fixes a bug where a translation would be executed too early.

If we translate upon loading the file we don't have a chance to load the translations before this point, resulting in a Jed error.

## Test instructions

* Load the term edit page in WordPress with yoast-components linked
* You shouldn't see an error now.

Fixes https://github.com/Yoast/wordpress-seo/issues/9779


igorschoester:
CR & Acceptance 👍

PR #548, opened 2018-05-23
Decode separator replacement variable

danielFangstrom:
## Summary

* Decodes the separator replacement variable before it is passed along to the renderer

* Finds the replacement variable `sep` and decodes the HTML code to a displayable symbol

## Test instructions

* Change the `sep` replacement variable to different symbols and check if they are displayed properly in the title and the description (no escaped HTML codes). The `sep` replacement variable can be changed in your local wp-admin sidebar by going to SEO > Search Appearance > General.

Fixes https://github.com/Yoast/wordpress-seo/issues/9787


igorschoester:
Like discussed in person. I think you should've hooked into the `processReplacementVariables` function instead of changing the replacement variables themselves. But let's not keep that from merging, as this works too!
CR & Acceptance 👍

PR #556, opened 2018-05-25
Set the meta description length to 156 characters

afercia:
## Summary

* Sets the snippet meta description length to 156 characters on both the desktop and mobile previews

- uses lodash `truncate` to truncate the string only where there are spaces, keeping the words not truncated
- adds a space and three dots ` ...` at the end of the string
- removes any height calculation as the meta description height is not based on the number of lines any longer

## Test instructions

- on yoast-components run `yarn start` then go to localhost:3333
- play with the meta description, remove, add words, start from an empty one, do all sort of things
- verify that on both the mobile and desktop preview the description gets truncated at 156
- worth noting the final result in the description preview can be shorter than 156, depending on the length of the last words (especially the second-last one): this is because the whole word gets removed

Fixes https://github.com/Yoast/wordpress-seo/issues/9837


boblinthorst:
CR and acceptance: code is fine, and works as described. I'll fix the merge conflicts and merge.

PR #562, opened 2018-05-28
Snippet preview title and description progress measurement overhaul

IreneStr:
## Summary

* Make the progress bars for the snippet title and description respond to the content of the input fields.
* Multiple spaces and spaces at the beginning and end of the snippet description are no longer taken into account anymore for the description length progress.
* Newlines in the snippet description are replaced by a space before being displayed in the snippet preview.

* Move the measurement of the title and description from `wordpress-seo` to `yoast-components` in order to make the progress bars work without a CMS-specific implementation.
* Remove the test sliders for the progress bars, because the progress bars now respond to the content of the input fields.
* Strip multiple spaces and spaces at the beginning and end of the description.
* I've made a distinction between `mapDataToPreview` and `mapDataToMeasurement`. Most of the mapping is taking place in `mapDataToMeasurement`. In `mapDataToPreview` is decided whether to use the hand-written description or, if that one is empty, the generated description. This distinction exists because the generated description shouldn't be measured because the progress bar should be empty when a generated description is being used.
* Although it is not possible anymore to add newlines to the description and title, the replacement of newlines by spaces is kept in this PR (I've discussed this with @atimmer) because sticking lines together without spaces is never a good idea. 

## Test instructions

* Link `stories/add-title-and-description-assessment` to `wordpress-seo`. Test whether the progress bars respond to the content of the input fields.
* Add multiple spaces to the description, both in the middle, at the beginning and the end. Multiple spaces shouldn't be counted for the description length, and spaces from the beginning and end should be stripped as well.
* Play around with changing the title and metadescription, reload the page etc. to make sure everything works as expected with mapping from input field to the preview and from the input field to the progress bars.

Fixes https://github.com/Yoast/wordpress-seo/issues/9777
Fixes https://github.com/Yoast/wordpress-seo/issues/9913

### DIFF
--- a/composites/Plugin/SnippetEditor/components/SnippetEditor.js
+++ b/composites/Plugin/SnippetEditor/components/SnippetEditor.js
@@ -2,6 +2,9 @@
 import React from "react";
 import PropTypes from "prop-types";
 import { __ } from "@wordpress/i18n";
+import MetaDescriptionLengthAssessment from "yoastseo/js/assessments/seo/metaDescriptionLengthAssessment";
+import PageTitleWidthAssesment from "yoastseo/js/assessments/seo/pageTitleWidthAssessment";
+import { measureTextWidth } from "yoastseo/js/helpers/createMeasurementElement";
 
 // Internal dependencies.
 import SnippetPreview from "../../SnippetPreview/components/SnippetPreview";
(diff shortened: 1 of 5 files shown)


composites/Plugin/SnippetEditor/components/SnippetEditor.js
```diff
@@ -87,6 +136,44 @@ class SnippetEditor extends React.Component {
 		this.setEditButtonRef = this.setEditButtonRef.bind( this );
 	}
 
+	/**
+	 * Returns whether the old and the new data are the same.
```
igorschoester: This description might be a tad confusing while still being logically correct 😉

igorschoester:
CR 👍

igorschoester:
Acceptance 🚧 
In the meta description ignored characters do count for an instance and is visible in the progressbar dancing back and forth when typing space for example.
test
Note: this does not seem to happen in wordpress-seo.

danielFangstrom:
CR 👍

danielFangstrom:
ACC 🚧 
When the meta description is empty, changes to the excerpt (which is now the fallback for the meta description) are only reflected in the meta description's progress bar when the title field is changed.
Also, there's code to ignore newlines which is now redundant.

igorschoester:
CR & Acceptance 👍

Issue #554, opened 2018-05-24
IE11: the replaced variables in the snippet editor fields disrupt keyboard navigation


afercia:
Noticed while working on #542. It happens also before the changes introduced there, but it gets more serious as there's potentially a "keyboard trap", meaning keyboard users can't tab away from the fields.

To reproduce, you need to test with IE 11.

Turns out the text that represents "replaced variables" in the snippet editor fields is actually focusable in IE 11. See in the screenshots below, that's what happens when tabbing within the field:

![screenshot 132](https://user-images.githubusercontent.com/1682452/40490842-f8db352c-5f6c-11e8-9bc6-342c3e9182c4.png)

![screenshot 133](https://user-images.githubusercontent.com/1682452/40490843-f8fbf08c-5f6c-11e8-94db-d50044c79ac3.png)

![screenshot 134](https://user-images.githubusercontent.com/1682452/40490844-f91d48a4-5f6c-11e8-8bd7-e774d50c7591.png)

As you can see, the replaced variables (actually they're text surrounded with some HTML elements) are focusable and  get a "resize handle" or just receive focus while tabbing within the field.

In some circumstances, sorry but I couldn't reproduce consistently, tabbing gets "stuck" on one of these elements and there's no way to tab away from the field.

I guess we should investigate why IE11 thinks these elements are focusable in the first place. It's very likely something related to draft.js and could be a known issue. In a way, this remembers me something that happened with SVG icons and was solved adding a HTML attribute `focusable="false"`.

afercia:
Inspired by this comment found by @IreneStr https://github.com/facebook/draft-js/issues/657#issuecomment-335770848

Seems any CSS property that triggers `hasLayout` in IE 11 in combination with `contenteditable` triggers the `resize handles`. 

Removing any CSS that triggers `hasLayout` fixes it and also fixes the strange behavior with left / right arrows noticed in https://github.com/Yoast/yoast-components/pull/568#issuecomment-395452346

This would require to reset `display: inline-block` on the "highlighted" words, which makes them look not so nice. Will try to research a bit further.

afercia:
It was decided to handle this in #568 as part of the fixes for the title field.

afercia:
#584 addresses only the resize-handle on the title-field.

Still to do in this issue: gif the resize-handles on the highlighted replaced variables and the left/right arrow keys behavior. To fix it, we should make the highlighted words not have `hasLayout` in IE 11 only.

PR #568, opened 2018-05-30
Add a button to insert a replacement variable in the snippet editor

### LINKED ISSUE #552: Snippet variables inserter
Add a button to insert snippet variables.

What the button basically does, is insert a `%` which will trigger the autocomplete. This also makes that behavior more discoverable. Designs coming up shortly by @uxkai

igorschoester:
## Summary

* N/A

* Adds the plus-circle path to the SvgIcon component.
* Removes the current selected text.
* Insert the trigger `%` and ensure the autocomplete gets shown by adding spaces around it as needed.
* Move the caret if needed and focus the editor.
* Copied the button style from the `EditSnippetButton`. And adapted to positioning.

## Test instructions

* Check thoroughly if you get the suggestions autocomplete every time you hit the button.
* If there is any text selected it should be removed.
* If the caret is surround by text it should make room around the % to make the autocomplete show.

Fixes #552


boblinthorst:
cr: some small comments, also merge conflicts

composites/Plugin/SnippetEditor/components/SnippetEditorFields.js
```diff
@@ -114,6 +116,28 @@ const SimulatedLabel = styled.div`
 	cursor: pointer;
 	font-size: 16px;
 	font-family: Arial, Roboto-Regular, HelveticaNeue, sans-serif;
+	margin-bottom: 5px;
+`;
+
+const InsertVariableButton = Button.extend`
+	height: 33px;
+	border: 1px solid #dbdbdb;
```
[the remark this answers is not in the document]
igorschoester: It was the issue to copy the style of the edit snippet button. The color is hardcoded like this there. And this color is not in colors.json.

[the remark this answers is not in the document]
IreneStr: Please investigate if you can use yoast-components' IconButton here. Please discuss with Anton why the edit snippet button isn't using IconButton either.

composites/Plugin/SnippetEditor/text.js
```diff
@@ -0,0 +1,126 @@
...
+
+/**
+ * Replaces current selection with text.
+ *
+ * @param {EditorState} editorState The Draft.js editor state.
+ * @param {string}      text        The text to insert.
+ *
+ * @returns {EditorState} The new editor state.
+ */
+export const replaceText = ( editorState, text ) => {
```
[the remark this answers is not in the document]
IreneStr: @boblinthorst ☘️

composites/Plugin/SnippetEditor/text.js
```diff
@@ -0,0 +1,114 @@
...
+ * @param {EditorState} editorState The Draft.js editor state.
+ * @param {string}      text        The text to insert.
+ *
+ * @returns {EditorState} The new editor state.
+ */
+export const insertText = ( editorState, text ) => {
+	const content = editorState.getCurrentContent();
+	const selection = editorState.getSelection();
+
+	// If there is a selection instead of a caret do nothing.
```
IreneStr: please add comma after 'caret' to improve readability

IreneStr: Also: this, in combination with the function name sounds like you can't insert text when you have something selected. Is that really the case?

composites/Plugin/SnippetEditor/text.js
```diff
@@ -0,0 +1,114 @@
...
+
+	const newContent = Modifier.removeRange( content, selection, "backward" );
+	return EditorState.push( editorState, newContent, "remove-range" );
+};
+
+/**
+ * Moves the caret to the index.
+ *
+ * @param {EditorState} editorState The Draft.js editor state,
+ * @param {int}         index       The index of the caret to be.
```
IreneStr: number

composites/Plugin/SnippetEditor/text.js
```diff
@@ -0,0 +1,114 @@
+import { EditorState, SelectionState, Modifier } from "draft-js";
```
IreneStr: Based on the file name, I have no idea what this file does

composites/Plugin/SnippetEditor/components/ReplacementVariableEditor.js
```diff
@@ -222,6 +253,43 @@ class ReplacementVariableEditor extends React.Component {
...
+	 * Ensuring the autocomplete shows by adding spaces around the trigger if needed.
+	 *
+	 * @returns {void}
+	 */
+	triggerReplacementVariableSuggestions() {
+		// First remove any selected text.
+		let editorState = removeSelectedText( this.state.editorState );
+
+		// Get the current block text.
+		const selection = editorState.getSelection();
```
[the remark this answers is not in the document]
igorschoester: Not selected text, but it still has the caret position. Which is on a certain line (block) in the text.

composites/Plugin/SnippetEditor/components/ReplacementVariableEditor.js
```diff
@@ -222,6 +253,43 @@ class ReplacementVariableEditor extends React.Component {
...
+		const content = editorState.getCurrentContent();
+		const text = getAnchorBlock( content, selection ).getText();
```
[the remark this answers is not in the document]
igorschoester: Content contains the text as well as the selection before and after rendering. But I realize this could be confusing so I'm going to rename all the content to contentState and selection to selectionState.

composites/Plugin/SnippetEditor/components/ReplacementVariableEditor.js
```diff
@@ -222,6 +253,43 @@ class ReplacementVariableEditor extends React.Component {
...
+		const index = getCaretOffset( selection );
+
+		// Determine the trigger text.
+		const needsPrefix = ! hasWhitespaceAt( text, index - 1 );
+		const needsSuffix = ! hasWhitespaceAt( text, index );
+		const trigger = getTrigger( needsPrefix, needsSuffix );
```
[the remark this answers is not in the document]
igorschoester: This is needed to make the suggestion popover appear.

composites/Plugin/SnippetEditor/components/SnippetEditorFields.js
```diff
@@ -57,7 +60,6 @@ const InputContainer = styled.div.attrs( {
 	position: relative;
 	font-family: Arial, Roboto-Regular, HelveticaNeue, sans-serif;
 	font-size: 14px;
-	margin-top: 5px;
```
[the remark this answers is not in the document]
igorschoester: Because the `TriggerReplacementVariableSuggestionsButton` needs less space between the input and itself. So I moved the margin to the label as margin-bottom.

composites/Plugin/SnippetEditor/components/SnippetEditorFields.js
```diff
@@ -356,6 +470,7 @@ SnippetEditorFields.defaultProps = {
 		actual: 0,
 		score: 0,
 	},
+	mobileWidth: 356,
```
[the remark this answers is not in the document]
igorschoester: Yes, this is meant to be able to adjust the magic number. But as you correctly pointed out. This should really be measured on the fly because we don't know how wide the label will be. Will create an issue to tackle this.

composites/Plugin/SnippetEditor/components/SnippetEditorFields.js
```diff
@@ -114,6 +116,28 @@ const SimulatedLabel = styled.div`
 	cursor: pointer;
 	font-size: 16px;
 	font-family: Arial, Roboto-Regular, HelveticaNeue, sans-serif;
+	margin-bottom: 5px;
+`;
+
+const TriggerReplacementVariableSuggestionsButton = Button.extend`
```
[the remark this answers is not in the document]
igorschoester: Will create another issue to change this plus the snippet editor button to IconButtons and change the color to a value from our style guide.

composites/Plugin/SnippetEditor/text.js
```diff
@@ -0,0 +1,114 @@
...
+ *
+ * @param {string} text  The text to get the character from.
+ * @param {int}    index The index of the character in the text.
+ *
+ * @returns {boolean} True if the character is out of bounds or whitespace.
+ */
+export const hasWhitespaceAt = ( text, index ) => {
+	const char = text.charAt( index );
+
+	// No character found equals whitespace.
```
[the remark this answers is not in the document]
igorschoester: Because of the way I use it with the trigger. I can not think of a better name. Suggestions welcome!

IreneStr:
CR 👍

afercia:
Pushed a few minor JSDocs improvements.

composites/Plugin/SnippetEditor/components/SnippetEditorFields.js
```diff
@@ -114,34 +117,55 @@ const SimulatedLabel = styled.div`
...
+	height: 33px;
+	border: 1px solid #dbdbdb;
+	box-shadow: none;
+	font-family: Arial, Roboto-Regular, HelveticaNeue, sans-serif;
```
afercia: Why are we using this font stack? WP uses native fonts, we should remove this. There are other occurrences though, will open a separate issue.

composites/Plugin/SnippetEditor/components/SnippetEditorFields.js
```diff
@@ -114,34 +117,55 @@ const SimulatedLabel = styled.div`
...
+	fill: ${ colors.$color_grey_dark };
+	padding-left: 8px;
+	float: right;
+	margin-top: -35px; // negative height + 2 for spacing
+
+	${ props => props.isSmallerThanMobileWidth && `
```
afercia: this should be `hasMobileWidth` (which is the passed prop) while `isSmallerThanMobileWidth` is the state property the passed prop depends on. Currently, when the viewport is less than the mobile width (356), there's no change in the CSS. I'd consider to rename `hasMobileWidth` to something more appropriate.

screen shot 2018-06-07 at 15 06 18

composites/Plugin/SnippetEditor/components/SnippetEditorFields.js
```diff
@@ -156,19 +180,37 @@ class SnippetEditorFields extends React.Component {
...
 		this.uniqueId = uniqueId( "snippet-editor-field-" );
 
+		this.state = {
+			isSmallerThanMobileWidth: false,
+		};
+
 		this.setRef = this.setRef.bind( this );
+		this.setEditorRef = this.setEditorRef.bind( this );
+		this.triggerReplacementVariableSuggestions = this.triggerReplacementVariableSuggestions.bind( this );
+		this.debouncedUpdateIsSmallerThanMobileWidth = debounce( this.updateIsSmallerThanMobileWidth.bind( this ), 500 );
```
afercia: Maybe a shorter debounce value could work, for example WordPress uses 200 for its `wp-window-resized` custom event. Any specific reason for 500?

composites/Plugin/SnippetEditor/components/SnippetEditorFields.js
```diff
@@ -199,6 +262,33 @@ class SnippetEditorFields extends React.Component {
...
+	 * Inserts a % into a ReplacementVariableEditor to trigger the replacement variable suggestions.
+	 *
+	 * @param {string} fieldName The field name to get the ref for.
+	 *
+	 * @returns {void}
+	 */
+	triggerReplacementVariableSuggestions( fieldName ) {
+		const element = this.elements[ fieldName ];
+
+		element.triggerReplacementVariableSuggestions();
```
afercia: I find very confusing this function has the same name of the one that comes from the ReplacementVariableEditor class. Any specific reason why they should have the same name?

composites/Plugin/SnippetEditor/replaceText.js
```diff
@@ -0,0 +1,115 @@
...
+ * @param {string}      text        The text to insert.
+ *
+ * @returns {EditorState} The new editor state.
+ */
+export const insertText = ( editorState, text ) => {
+	const contentState = editorState.getCurrentContent();
+	const selectionState = editorState.getSelection();
+
+	// If there is a selection instead of a caret do nothing.
+	if ( ! selectionState.isCollapsed() ) {
```
afercia: Seems to me`selectionState.isCollapsed()` is always `true` whether there's a caret or a selection, so this condition never passes. Not sure why, also `document.getSelection()` tells me it's always `collapsed`. Maybe there's something going on before this condition that turns a selection to a caret?

composites/Plugin/SnippetEditor/replaceText.js
```diff
@@ -0,0 +1,115 @@
...
+	// Default to the block where the anchor is currently at.
+	if ( blockKey === "" ) {
+		blockKey = getAnchorBlock( contentState, selectionState ).getKey();
+	}
+
+	const newSelectionState = SelectionState.createEmpty( blockKey )
+	                                   .merge( {
+		                                   anchorOffset: caretIndex,
+		                                   focusOffset: caretIndex,
+	                                   } );
```
afercia: Ideally, we should avoid this kind of indentation made of spaces or mixes of tabs and spaces.

afercia:
CR 🚧 

Noticed something buggy using Firefox, happens only on the title field:
- remove the default content from the title field
- enter something, e.g. `hello`
- press spacebar to enter a new word
- at this point, the caret is moved at the beginning of the field: expected: should stay at the end
- using the arrow keys, move the caret at the end
- notice there is a space
- type one more `hello`
- the space is removed, the resulting text is `hellohello` without a space

Doesn't happen in Chrome.

Also, please see the other comments.

afercia:
🚨 Just noticed something serious happens on Safari:

- remove the default content from the title field
- type `Hello %t` and choose `title` from the suggestions
- `title` is added and highlighted
- press Spacebar once: nothing happens
- press Spacebar again: the app crashes, page goes blank, see error in the console in the screenshot below:

screen shot 2018-06-07 at 15 43 19

composites/Plugin/SnippetEditor/components/SnippetEditorFields.js
igorschoester: This was due to it having to match the old style for the color. I assume the same counts for the font stack. New issue sounds nice! On my todo is creating a issue to replace this with an IconButton.

composites/Plugin/SnippetEditor/components/SnippetEditorFields.js
igorschoester: No specific reason, will change to 200.

composites/Plugin/SnippetEditor/components/SnippetEditorFields.js
igorschoester: Forgot this in my rename to `isSmallerThanMobileWidth`, will fix. Good spot!

composites/Plugin/SnippetEditor/replaceText.js
```diff
@@ -0,0 +1,115 @@
...
+ * @param {string} text  The text to get the character from.
+ * @param {number} index The index of the character in the text.
+ *
+ * @returns {boolean} True if the character is out of bounds or whitespace.
+ */
+export const hasWhitespaceAt = ( text, index ) => {
+	const char = text.charAt( index );
+
+	// No character found equals whitespace.
+	if ( char.length === 0 ) {
```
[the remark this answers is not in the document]
igorschoester: I don't think that would be easier to read. Converting a number to a boolean and then inverting it.

composites/Plugin/SnippetEditor/replaceText.js
igorschoester: In `triggerReplacementVariableSuggestions` it removes the selected text before. So in the current use-case what you are saying is true. But when you use the `insertText` without that check it would error if it is not collapsed.

IreneStr:
@afercia Is the bottom one also only in the title field?

igorschoester:
Both bugs you found are not related to this PR it seems 😞

afercia:
Testing on Windows: Edge

- remove the default content from the title field
- as soon as I type a new character, the page goes blank (see in the screenshoot below the same error happens multiple times and React tries to remount till it gives up)

screenshot 20

screenshot 21

afercia:
Testing on Windows: IE 11

- remove the default content from the title field
- type something followed by a percentage sign
- the caret gets moved before the percentage sign, thus no suggestions are triggered (see screenshot below)
- expected: caret should stay where it is and suggestions should appear
- generally, typing in the title field is sluggish and becomes extremely sluggish as the number of words increases; also deleting with backspace is slowww and very sluggish
- pressing left and arrow keys in the title field moves the caret at the beginning and the end of the field: expected: should move by one character

screenshot 19

composites/Plugin/SnippetEditor/components/SnippetEditorFields.js
afercia: Hm no great ideas but I'd say the one that actually triggers the suggestions is the one from the ReplacementVariableEditor class? Instead, this one gets a field name an passes it to the other one.

igorschoester:
@afercia This PR seems to fix those bugs you found: https://github.com/Yoast/yoast-components/pull/584
I did not test in IE and Edge though.

afercia:
> I did not test in IE and Edge though.

I can do that :)

IreneStr:
CR done 👍

afercia:
Re: the strange behavior of the left / right arrows in IE 11, seems the infamous IE property `hasLayout` is responsible for that. See https://github.com/Yoast/yoast-components/issues/554#issuecomment-395692626

IreneStr:
I've opened two non-blocking issues:
https://github.com/Yoast/yoast-components/issues/586
https://github.com/Yoast/yoast-components/issues/587

PR #569, opened 2018-06-01
Only suggest replacevars starting with the search value.

boblinthorst:
## Summary

* N/A

* Stopped using draftjs-mention's own default filter, and started using a self-made filter that has no limit on matches, and that only matches from the start.

## Test instructions

* Test if the tests pass.
* Go to a replacevarseditorfield ( the title field of the snippet editor for instance)
* Type a `%` and check if all available replacevars are suggested.
* Type a character, and check if only those replacevars are suggested that start with that character.

Fixes #https://github.com/Yoast/wordpress-seo/issues/9891


maartenleenders:
Acceptance 👍

PR #570, opened 2018-06-04
9892 remove prefixing hyphens

boblinthorst:
## Summary

* N/A

* Used if statement to filter the dashes. Changed what is sent to the function to only be the slug. This way, It's easier to check for the start of what the user has typed. Otherwise, the baseurl also needed to be passed, which would've been inconvenient.

## Test instructions

* Check if the unit tests pass on wordpress-seo.
* Checkout this branch and it's sibling branch: https://github.com/Yoast/wordpress-seo/pull/9937
* Visit a post page, add a url with a prefix space and a trailing space, verify no dashes are shown in the snippet preview.

Hyphens typed by the user will unfortunately also be filtered, this has been discussed, and is not considered a big enough problem to stop this pull.

Fixes #https://github.com/Yoast/wordpress-seo/issues/9892


boblinthorst:
Replaced for https://github.com/Yoast/yoast-components/pull/571

PR #571, opened 2018-06-04
Remove leading spaces being changed to hyphen

boblinthorst:
## Summary

* N/A

* Used if statement to filter the dashes. Changed what is sent to the function to only be the slug. The baseurl after the protocol is removed has been sent along in a context variable. To be able to check where the user submitted slug begins.

## Test instructions

* Check if the unit tests pass on wordpress-seo.
* Checkout this branch and it's sibling branch: https://github.com/Yoast/wordpress-seo/pull/9937
* Visit a post page, add a url with a prefix space and a trailing space, verify no dashes are shown in the snippet preview.

Hyphens typed by the user will unfortunately also be filtered, this has been discussed, and is not considered a big enough problem to stop this pull.

Fixes #https://github.com/Yoast/wordpress-seo/issues/9892


danielFangstrom:
CR 👍

composites/Plugin/SnippetEditor/components/SnippetEditor.js
```diff
@@ -352,17 +352,19 @@ class SnippetEditor extends React.Component {
...
 			title: this.processReplacementVariables( originalData.title ),
-			url: baseUrl.replace( "https://", "" ) + originalData.slug,
+			url: originalData.slug,
 			description: description,
 		};
 
 		// The mapping by the passed mapping function should happen before measuring.
 		if ( mapDataToPreview ) {
-			return mapDataToPreview( mappedData, originalData );
+			mappedData =  mapDataToPreview( mappedData, originalData );
```
IreneStr: I'd like to doublecheck this decision with @atimmer. Is it okay to have only the slug without the base URL at the point where the passed `mapDataToPreview` function is going to do its work? Can we assume the passed `mapDataToPreview` function will never need to change the base URL? And if so, calling it 'url' at this point might make things unclear.

atimmer: Please pass the `baseUrl` in a `context` 3th argument. Its a clean solution and makes it future compatible.

composites/Plugin/SnippetEditor/components/SnippetEditor.js
```diff
@@ -352,17 +352,24 @@ class SnippetEditor extends React.Component {
...
 		};
 
+		const context = {
+			shortenedBaseUrl,
+		};
+
 		// The mapping by the passed mapping function should happen before measuring.
 		if ( mapDataToPreview ) {
-			return mapDataToPreview( mappedData, originalData );
+			return mapDataToPreview( mappedData, context );
```
afercia: I find confusing that this `mapDataToPreview` prop (which passes a function) has the same name of the `mapDataToPreview` function in the class itself. Also, it gets passed two arguments but the passed mapping function is `mapEditorDataToPreview` and it accepts one argument, unless I'm missing something.

afercia:
CR 🚧 just because I have a couple question. Very likely because I don't know enough about this feature, sorry 🙂

composites/Plugin/SnippetEditor/components/SnippetEditor.js
boblinthorst: You're completely right, this one is my mistake, I forgot to push changes on the worpdpress-seo sibling pr 😓

composites/Plugin/SnippetEditor/components/SnippetEditor.js
```diff
@@ -352,17 +352,24 @@ class SnippetEditor extends React.Component {
 		// Strip multiple spaces and spaces at the beginning and end.
 		description = stripSpaces( description );
 
-		const mappedData = {
+		const shortenedBaseUrl = baseUrl.replace( "https://", "" );
```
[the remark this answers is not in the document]
boblinthorst: I'm not the one who made that originally, so I'll ask Anton.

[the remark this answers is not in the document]
atimmer: Oh yeah, Google shows `http://` in the search results. But doesn't show `http://`.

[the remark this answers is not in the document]
atimmer: Ok, so we've actually implemented this incorrectly.

It is the other way around: https://github.com/Yoast/YoastSEO.js/blob/develop/src/snippetPreview.js#L623-L624. I did a quick Google search to verify this.

composites/Plugin/SnippetEditor/components/SnippetEditor.js
boblinthorst: I've discussed the name of the function with Irene. Unfortunately, both names describe exactly what they do, so renaming it quickly makes the new name suboptimal. 

If the reviewer has a better suggestion for the name, you're welcome to suggest them, but we couldn't think of one.

afercia: I'd suggest to rename the prop that passes the function to start with, currently it's:
`mapDataToPreview={ mapEditorDataToPreview }`
and rename it to
`mapEditorDataToPreview={ mapEditorDataToPreview }`

This would require a small change also in the plugin. Alternatively, not destructure the props and use `this.props.mapDataToPreview` could make the code more readable.

composites/Plugin/SnippetEditor/components/SnippetEditor.js
```diff
@@ -385,24 +385,31 @@ class SnippetEditor extends React.Component {
...
-		const { baseUrl, mapDataToPreview } = this.props;
+		const { baseUrl, mapEditorDataToPreview } = this.props;
 
 		let description = this.processReplacementVariables( originalData.description );
 
 		// Strip multiple spaces and spaces at the beginning and end.
 		description = stripSpaces( description );
 
-		const mappedData = {
+		const shortenedBaseUrl = baseUrl.replace( /http:\/\//ig, "" );
```
igorschoester: A minor point: The base url should only ever start with `http://` and never contain more in the middle somewhere. We could optimize the regex to:
```
baseUrl.replace( /^http:\/\//i, "" );
```

composites/Plugin/SnippetEditor/components/SnippetEditor.js
```diff
@@ -112,7 +112,7 @@ class SnippetEditor extends React.Component {
 	 * @param {Object} props.descriptionLengthProgress The values for the
 	 *                                                   description length
 	 *                                                   assessment.
-	 * @param {Function} props.mapDataToPreview          Function to map the editor
+	 * @param {Function} props.mapEditorDataToPreview          Function to map the editor
```
igorschoester: With renaming the docs alignment went awry 😱 😉

igorschoester:
CR & Acceptance 👍

PR #572, opened 2018-06-04
Stories/9834 ditch snippet variables

danielFangstrom:
## Summary

* Removes certain replacement variables from appearing in the suggestions in the snippet editor (`currentdate`, `currenttime`, `currentyear`, `currentmonth`, `currentday`, `userid`, `searchphrase`, and `archive_title`)

* Introduced a new prop which can contain the names of any replacement variables that should be removed from the suggestions
* This allows specific inclusion/exclusion of replacement variables depending on the instance of the replacement variable editor (the upcoming implementation for the settings page can include `searchphrase` for example)
* This also allows the replacement variables to be used as normal regardless of whether or not they appear in the suggestions menu

## Test instructions

* Link wordpress-seo to yoast-components and build
* When editing a snippet, bring up the replacement variable suggestions in the title or meta description fields
* Make sure that the above mentioned replacement variables are not listed in the suggestions 
* Make sure that's the case for both pages and posts
* Typing any of the above replacement variables in their full form (e.g. %%currentdate%%) will correctly display the value (but not bring it up in the suggestions) 

Fixes https://github.com/Yoast/wordpress-seo/issues/9834

### DIFF
--- a/composites/Plugin/SnippetEditor/components/ReplacementVariableEditor.js
+++ b/composites/Plugin/SnippetEditor/components/ReplacementVariableEditor.js
@@ -67,8 +67,8 @@ class ReplacementVariableEditor extends React.Component {
 	constructor( props ) {
 		super( props );
 
-		const { content: rawContent, excludeReplaceVars } = this.props;
-		const replacementVariables = this.excludeReplaceVars( this.props.replacementVariables, excludeReplaceVars );
+		const { content: rawContent } = this.props;
+		const replacementVariables = this.excludeReplaceVars( this.props.replacementVariables, this.props.excludeReplaceVars );
 		const unserialized = unserializeEditor( rawContent, replacementVariables );
 
 		this.state = {
--- a/composites/Plugin/SnippetEditor/components/ReplacementVariableEditor.js
+++ b/composites/Plugin/SnippetEditor/components/ReplacementVariableEditor.js
@@ -67,7 +67,8 @@ class ReplacementVariableEditor extends React.Component {
 	constructor( props ) {
 		super( props );
 
-		const { content: rawContent, replacementVariables } = this.props;
+		const { content: rawContent, excludeReplaceVars } = this.props;
+		const replacementVariables = this.modifyReplaceVars( this.props.replacementVariables, excludeReplaceVars );
 		const unserialized = unserializeEditor( rawContent, replacementVariables );
 
 		this.state = {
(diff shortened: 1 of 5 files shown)


composites/Plugin/SnippetEditor/components/SnippetEditor.js
```diff
@@ -528,6 +531,10 @@ SnippetEditor.defaultProps = {
 	mapDataToPreview: null,
 	generatedDescription: "",
 	locale: "en",
+	excludeReplaceVars: [
```
maartenleenders: In this case, setting a defaultProp that is filled with values has the unintuitive consequence that one needs to explicitly pass `excludeReplaceVars` as an empty array to receive all replaceVars, rather than omitting the prop.

I guess the alternative (having to pass these 8 values in a string almost everytime) is not desirable either though... Can't we set the values to be excluded in each contect in a look-up helper object somewhere? For example
```{
    edit-post: {
        context: "edit-post",
        excludeReplaceVars: [ "your_stuff, "my_stuff", ]
    },
    edit-archive: {
        context: "edit-archive",
        excludeReplaceVars: [ "your_archive_stuff ],
    },
}
```
Etc.? Can't oversee right now whether that would be desirable.

composites/Plugin/SnippetEditor/components/ReplacementVariableEditor.js
```diff
@@ -67,7 +67,8 @@ class ReplacementVariableEditor extends React.Component {
 	constructor( props ) {
 		super( props );
 
-		const { content: rawContent, replacementVariables } = this.props;
+		const { content: rawContent, excludeReplaceVars } = this.props;
+		const replacementVariables = this.excludeReplaceVars( this.props.replacementVariables, excludeReplaceVars );
```
maartenleenders: You have removed `replacementVariables` being deconstructed from `this.props`, and call `this.props.replacementVariables`, yet include deconstruction of `excludeReplaceVars`. Nitpicky thing, but would be nice if the design patterns here are consistent.

maartenleenders:
CR 🚧 minor things. Questions, really.

composites/Plugin/SnippetEditor/components/SnippetEditor.js
danielFangstrom: After the discussion we had, and with the new instances of the replacement variable editor having a separate architecture, we will keep the excluded variables in the default props. The look-up helper object is a really nice idea though, and is something to keep in mind.

IreneStr:
Acceptance 👍 
Note: None of the mentioned replacevars show up in the replace vars, as expected. However, userid, searchphrase, and archive_title were unparsed to begin with, and remain unparsed if you use them (= no regression).

Issue #574, opened 2018-06-04
Edge and IE11: standalone pages render only with the console open


afercia:
Noticed while testing https://github.com/Yoast/yoast-components/pull/573

On Windows, using Edge or IE11, the standalone examples at `localhost:3333` render a blank page.

Only when opening the Edge and IE 11 "dev tools", the app renders.

A first problem might be the various `console.*` used here and there (as far as I remember on IE 11 `console` is available only with the dev tools open). However, I guess that's not the only problem.

Worth noting this makes testing on these two browsers far from ideal.

afercia:
After I've successfully (and painfully) updated Windows, my Edge version is now:
Microsoft Edge 42.17134.1.0
Microsoft EdgeHTML 17.17134

and seems there's no need to open the console any longer.

IE 11 still needs the console open to actually see something, otherwise there's just a blank page.

PR #584, opened 2018-06-07
Fix single line browser compatibility

igorschoester:
## Summary

* N/A

* Remove white-space nowrap style on the TitleInputContainer in the SnippetEditorFields. This was creating problems in other browsers than Chrome.

## Test instructions

* https://github.com/Yoast/yoast-components/pull/568#issuecomment-395423015
* https://github.com/Yoast/yoast-components/pull/568#issuecomment-395426406
* https://github.com/Yoast/yoast-components/pull/568#issuecomment-395450196
* https://github.com/Yoast/yoast-components/pull/568#issuecomment-395452346

## Note

I did not test in IE and Edge to see if this fixes those yet.

### DIFF
--- a/composites/Plugin/SnippetEditor/components/SnippetEditorFields.js
+++ b/composites/Plugin/SnippetEditor/components/SnippetEditorFields.js
@@ -78,7 +78,6 @@ const TitleInputContainer = InputContainer.extend`
 		line-height: 24px;
 		height: 24px;
 		overflow: hidden;
-		white-space: nowrap;
 	}
 `;
 
--- a/composites/Plugin/SnippetEditor/components/SnippetEditorFields.js
+++ b/composites/Plugin/SnippetEditor/components/SnippetEditorFields.js
@@ -75,9 +75,8 @@ const InputContainer = styled.div.attrs( {
 
 const TitleInputContainer = InputContainer.extend`
 	.public-DraftStyleDefault-block {
+		// Don't use properties that trigger hasLayout in IE11.
 		line-height: 24px;
-		height: 24px;
-		overflow: hidden;
 	}
 `;
 
--- a/composites/Plugin/SnippetEditor/tests/__snapshots__/SnippetEditorTest.js.snap
+++ b/composites/Plugin/SnippetEditor/tests/__snapshots__/SnippetEditorTest.js.snap
@@ -995,8 +995,6 @@ exports[`SnippetEditor calls callbacks when the editors are focused or changed 1
 
 .c32 .public-DraftStyleDefault-block {
   line-height: 24px;
-  height: 24px;
-  overflow: hidden;
 }
 
 .c34 {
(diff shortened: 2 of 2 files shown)


afercia:
Tested on all major browser. The biggest issue is that removing `white-space: nowrap;` makes the title field behave like a textarea again. Just paste some very long content (for example, use the meta description one) and see the content goes in multiple lines.

That said:
- it solves the crash in Safari https://github.com/Yoast/yoast-components/pull/568#issuecomment-395426406
- solves the Firefox issues https://github.com/Yoast/yoast-components/pull/568#issuecomment-395423015
- solves the IE11 issues https://github.com/Yoast/yoast-components/pull/568#issuecomment-395452346 except `pressing left and arrow keys in the title field moves the caret at the beginning and the end of the field: expected: should move by one character`
- partially solves the crash in Edge https://github.com/Yoast/yoast-components/pull/568#issuecomment-395450196 it still crashes when:
  - remove the default content from the title field
  - type `Hello %t` and select "title" from the suggestions
  - at this point the replacevar is correctly inserted and highlighted
  - press Ctrl+A to select all content
  - with the content still selected, type something: crash 

With regards to the very sluggish typing responsiveness in Edge and IE, it still happens at a point that makes it almost unusable especially with very long content. Will open a separate issue.

Worth noting that to test in Edge and IE there's still the need to open the browser's console as noted in  #574

afercia:
FWIW the error I get in Safari seems to be this same error: https://github.com/Automattic/wp-calypso/issues/8047 worth investigating.

In Edge, the error (a different error) persists, I've tried to:
- remove the single line plugin: still crashes
- remove the mentions plugin: works

afercia:
Re: the Edge issue, we were able to reproduce it on the Mentions plugin online demo and reported the issue upstream: https://github.com/draft-js-plugins/draft-js-plugins/issues/1124

afercia:
It was decided to not make the title field look like a (single line) input field to removing `white-space: nowrap` si good but then we should also remove the height.

This gives us the opportunity to fix another buggy behavior in IE 11 which shows the "resize handles" round the field. We should remove any CSS property that triggers `hasLayout` (also from the highlighted text).

IreneStr:
@afercia Could you please add test instructions which specify what bugs in which browser should be fixed in this PR?

boblinthorst:
CR: I don't have the background you have built up yesterday on this issue. So I'll trust removing those props fixes that. Code seems fine otherwise. Acceptance test is more important for this PR anyway, so that should be done carefully.

afercia:
For reference, here's the bugs fixed by this PR. Please follow the links for instructions on how to reproduce the bugs:


The removal of `white-space: nowrap;` fixes:
- crash in Safari https://github.com/Yoast/yoast-components/pull/568#issuecomment-395426406
- caret behavior in Firefox https://github.com/Yoast/yoast-components/pull/568#issuecomment-395423015

The removal of CSS properties that trigger `hasLayout` in IE 11 like `height: 24px;` and `overflow: hidden;` fixes:
- resize handles appearing on the title field in IE 11 https://github.com/Yoast/yoast-components/issues/554#issuecomment-395692626

Still to fix in IE11: resize handles on the highlighted words and keyboard trap, will be addressed in #554
Still to fix in Edge: app crashes, see #591

IreneStr:
@afercia just confirmed the resize handles are gone as well, so let's merge! 🎉

PR #596, opened 2018-06-12
Use replacevar labels (nicenames) in suggestions

### LINKED ISSUE #577: Use snippet variables nicenames when selecting them in the mentions plugin
In https://github.com/Yoast/yoast-components/pull/566, Nicenames are introduced for snippet variables. But these nicenames weren't used when selecting a snippet variable in the mentions plugin. So this needs to be added.

yoast_components-2

Needs https://github.com/Yoast/wordpress-seo/pull/10065
Needs https://github.com/Yoast/yoast-components/issues/577

maartenleenders:
## Summary

* Replacevar suggestions now use nicenames rather than raw replacevariable names.

## Test instructions

* Test whether the dropdown suggestion list (after typing `%`) uses nicenames, and will fill the nicename into the editor field.
* Verify that the preview still replaces the variable with the correct value.
* Save and load posts, and enter replacevars by just typing `%%title%%` etc., and see that that functionality also still works (so without using the mention/suggestions).

Fixes #577 
Requires https://github.com/Yoast/wordpress-seo/pull/10065 (for custom field replacevars)

### DIFF
--- a/composites/Plugin/SnippetEditor/components/ReplacementVariableEditor.js
+++ b/composites/Plugin/SnippetEditor/components/ReplacementVariableEditor.js
@@ -331,6 +331,13 @@ class ReplacementVariableEditor extends React.Component {
 		const { MentionSuggestions } = this.mentionsPlugin;
 		const { onFocus, onBlur, ariaLabelledBy, descriptionEditorFieldPlaceholder } = this.props;
 		const { editorState, replacementVariables } = this.state;
+		const replaceVarSuggestions = replacementVariables.map( ( variable ) => {
+			return {
+				...variable,
+				name: variable.label,
+				replaceName: variable.name,
+			};
+		} );
 
 		return (
 			<React.Fragment>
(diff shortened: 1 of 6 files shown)


danielFangstrom:
CR 👍

danielFangstrom:
ACC 🚧 
When typing a non-letter character after the percentage sign, the suggestions keep popping up, regardless of where the cursor is. Nothing but selecting an option from the suggestions will stop it from popping up.

maartenleenders:
The suggestions that kept coming up was because the Mention plugin doesn't like the suggestions being tampered with in the render function. I don't know why. @igorschoester suggested it might have to do with changing pointers.

We moved suggestion management to the state, and manage that state in the `onSearchChange` and `componentWillReceiveProps`. This pleased the mention plugin ;).

boblinthorst:
CR: ok 👍

maartenleenders:
@CarolineGeven yes, that is specifically what this PR fixes. See https://github.com/Yoast/yoast-components/issues/577

danielFangstrom:
ACC 🚧 
Custom fields with spaces in the name creates duplicate entries in the suggestions, which in turn causes React to throw a warning.

maartenleenders:
Fixed in https://github.com/Yoast/wordpress-seo/pull/10065

danielFangstrom:
ACC 🚧 (because of related PR)

maartenleenders:
Related PR issues will be fixed in https://github.com/Yoast/wordpress-seo/issues/10098

danielFangstrom:
ACC 👍

PR #597, opened 2018-06-12
Fix progress bar for empty meta description

danielFangstrom:
## Summary

* Only add date to the length of the description if the description is set.

* Added an additional check so that date is not accounted for in the progress bar in cases of empty descriptions
* This is a bug fix and not directly related to the connected issue 

## Test instructions

* In settings, make sure that Date in Snippet Preview is enabled
* Empty the meta description field
* The progress bar should be completely empty
* Typing a single character in the meta description should immediately take the length of the date into account

Fixes https://github.com/Yoast/wordpress-seo/issues/9985


igorschoester:
CR 👍

boblinthorst:
Acceptance 👍

PR #600, opened 2018-06-15
9798 clickability meta description field

danielFangstrom:
## Summary

* N/A

* Passes the OnBlur prop from the snippetEditor to the ReplacementVariableEditorStandalone, causing a blur event to now call the unsetFieldFocus function. This means that the active field is now correctly updated and the focus is correctly set on a click event.

## Test instructions

* Edit a post/page, and edit snippet
* Click on the meta description field, underneath the part containing the text
* The field should be selected, with the text cursor being at the last character in the text
* Try different combinations (first clicking the focus keyword field before clicking the description field, first clicking the title, etc.), all of which should produce the above behavior
* Try the same in the settings page

Fixes https://github.com/Yoast/wordpress-seo/issues/9798


afercia:
CR done 👍 

Noticed a couple pre-existing issues related to ARIA attributes, will open a separate issue.

Xyfi:
Acceptance :+1:

PR #604, opened 2018-06-19
Higher z-index for the mention plugin suggestions

igorschoester:
## Summary

* Fixes an issue where in cases with a lot of snippet variables it would sometimes be partially hidden behind the WordPress toolbar.

* Increased the z-index by 1000, because this is the value of the element it was behind has as z-index.

## Test instructions

* See issue.

Fixes https://github.com/Yoast/wordpress-seo/issues/10045


danielFangstrom:
ACC 👍

PR #605, opened 2018-06-19
Introduce an ErrorBoundary component and use it around the SnippetEditorFields

### LINKED ISSUE #591: Use error boundaries for the Snippet editor replacement (mentions plugin)
Splitting this out from #584.

Turns out there's a bug in the Mentions plugin that makes Microsoft Edge crash the snippet editor app in some edge cases. The issue was reported upstream, see https://github.com/draft-js-plugins/draft-js-plugins/issues/1124

In the meantime, we should try to use error boundaries to workaround this.

afercia:
## Summary

* Introduce an ErrorBoundary component to be used for the Snippet Editor.

- for now, the error boundary wraps the SnippetEditorFields component
- it can be changed to wrap any other components tree, if desired
- uses `a11ySpeak` to deliver an audible message to screen reader users

Design feedback needed /Cc @hedgefield 

We can't predict where we'll need to re-use this component and show errors, so we can't set arbitrary paddings / margins. For this reason, I've aligned the text to the centre and used a bottom border, as it seemed the most pragmatical solution to me. Of course, this can be changed if desired, just let me know. Personally, I'd add `moar` white space. Screenshot:

screen shot 2018-06-19 at 17 32 52

## Test instructions
You can test the standalone version or yarn-link to the plugin:

- with Microsoft Edge on Windows: select the title field content using Ctrl+A
- with the content still selected, press any character key to trigger the Edge error
- check the error message is correctly displayed
- an audible message is sent to the ARIA live region

If you can't test with Edge, you have to change the error boundary initial state to `true`, then open the snippet editor to see the error message.

Fixes #591

### DIFF
--- a/composites/basic/ErrorBoundary.js
+++ b/composites/basic/ErrorBoundary.js
@@ -46,7 +46,7 @@ export default class ErrorBoundary extends React.Component {
 	render() {
 		if ( this.state.hasError ) {
 			// Render any custom fallback UI.
-			const errorMessage = __( "Something went wrong, please reload the page.", "yoast-components" );
+			const errorMessage = __( "Something went wrong. Please reload the page.", "yoast-components" );
 			a11ySpeak( errorMessage, "assertive" );
 
 			return <ErrorContainer>{ errorMessage }</ErrorContainer>;
--- a/composites/basic/ErrorBoundary.js
+++ b/composites/basic/ErrorBoundary.js
@@ -1,0 +1,61 @@
+// External dependencies.
+import React from "react";
+import PropTypes from "prop-types";
+import styled from "styled-components";
+import { __ } from "@wordpress/i18n";
+import { speak as a11ySpeak } from "@wordpress/a11y";
+
+/* Internal dependencies */
+import colors from "../../style-guide/colors";
+
+const ErrorContainer = styled.p`
+	text-align: center;
+	margin: 0 0 16px;
+	padding: 16px 16px 8px 16px;
+	border-bottom: 4px solid ${ colors.$color_bad };
+	background: ${ colors.$color_white };
+`;
+
+
+export default class ErrorBoundary extends React.Component {
+	/**
+	 * Constructs the component.
+	 *
+	 * @param {Object} props The props for the component.
+	 */
+	constructor( props ) {
+		super( props );
+
+		this.state = { hasError: false };
+	}
+
+	/**
+	 * Catches errors in a components tree.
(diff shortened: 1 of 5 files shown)


igorschoester:
Acceptance 🚧 

I did not get the error on Windows 10
Microsoft Edge 41.16299.492.0
Microsoft EdgeHTML 16.16299

I did notice this issue: https://github.com/Yoast/wordpress-seo/issues/10087

When putting the initial state to true on Chrome I got these errors:
errors

afercia:
> When putting the initial state to true on Chrome I got these errors:

I get this error only when clicking on the preview (title, slug, or description)

I'd say it's in a way unrelated because when the error boundary renders the alternate UI (the error message), then there's no Editor rendered. As I happened to say a few times previously, a best practice is to always check the existence of an object or a DOM element before trying to use methods or get properties. In this case, `focusOnActiveFieldChange()` tries to use the `focus()` method on an element that doesn't exist at that point.

I see two ways to "fix" this:
- check for the existence of the DOM node before trying to use `activeElement.focus();` however there are probably other edge cases we may face
- wrap the error boundary around the whole UI including the snippet preview

Please do let me know what the preferred option is. /Cc @atimmer 
Also, still pending design feedback. /Cc @hedgefield

composites/basic/ErrorBoundary.js
```diff
@@ -0,0 +1,61 @@
...
+
+	/**
+	 * Renders the Error Boundary.
+	 *
+	 * @returns {ReactElement} The Error Boundary or its children.
+	 */
+	render() {
+		if ( this.state.hasError ) {
+			// Render any custom fallback UI.
+			const errorMessage = __( "Something went wrong, please reload the page.", "yoast-components" );
```
IreneStr: "Something went wrong. Please reload the page." 😉

IreneStr: "Something went wrong. Please reload the page." 😉

afercia:
In the latest commits:
- wrapped the error boundary around the whole editor, this means in case of an error also the snippet preview won't be rendered
- added check for the existence of the input fields elements before trying to use `focus()` on them
- added the error boundary in the Settings pages template editors
- improved the error message text

Screenshots of how it looks in the plugin:

screen shot 2018-06-20 at 14 58 27

screen shot 2018-06-20 at 14 59 06

Xyfi:
CR :+1: Acceptance :+1:

PR #610, opened 2018-06-20
Pass updated replacementvariables when measuring progress

maartenleenders:
## Summary

* Changes in replacevariables now cause progress in SEO Title and description to be measured with the new replacevariable values.

* I chose to optionally pass replacementVariables to `mapDataToMeasurements` (and eventually `processReplacementVariables`), so that when mapDataToMeasurements is accessed from the `componentWillReceiveProps` function, the "next" replacementVariables can be used.

## Test instructions

* Put `%%title%%` in the SEO Title input field.
* Change the title of the post. See that the progress changes accordingly.
* Do the same for the description (put `%%title%%` in it and change the title).
* It should work for any replacement variable that you change, and that is placed in the title or description field.

Fixes https://github.com/Yoast/wordpress-seo/issues/10047

### DIFF
--- a/composites/Plugin/SnippetEditor/components/SnippetEditor.js
+++ b/composites/Plugin/SnippetEditor/components/SnippetEditor.js
@@ -7,6 +7,7 @@ import PageTitleWidthAssesment from "yoastseo/js/assessments/seo/pageTitleWidthA
 import { measureTextWidth } from "yoastseo/js/helpers/createMeasurementElement";
 import stripSpaces from "yoastseo/js/stringProcessing/stripSpaces";
 import noop from "lodash/noop";
+import toJSON from "lodash/toJSON";
 
 // Internal dependencies.
 import SnippetPreview from "../../SnippetPreview/components/SnippetPreview";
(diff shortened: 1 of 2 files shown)


composites/Plugin/SnippetEditor/components/SnippetEditor.js
```diff
@@ -146,21 +146,27 @@ class SnippetEditor extends React.Component {
...
+	 * Returns whether the old and the new data or replacement variables are different.
 	 *
-	 * @param {Object} oldData The old data.
-	 * @param {Object} newData The new data.
+	 * @param {Object} prev The old props.
+	 * @param {Object} next The new props.
 	 * @returns {boolean} True if any of the data points has changed.
 	 */
-	shallowCompareData( oldData, newData ) {
+	shallowCompareData( prev, next ) {
```
atimmer: I meant to call this `prevProps` and `nextProps`.

boblinthorst:
CR: ok 👍

Xyfi:
Acceptance :construction: 

If you paste a title it now updates properly, but when you remove the entire title when selecting it and deleting it, it does not. Also when holding backspace until the title field is empty the progress bar jumps from empty to filled when the last character is removed from the title field.

title-measurement-error

danielFangstrom:
The issue is that when the value of the replacement variable is empty, it gets the `%%title%%` as a replacement. This takes the value from rawdata in the app, which is not updated by the same events. This is probably not the main issue though as the value is later correctly updated, but the progress bar is not updated. There probably needs to be an additional trigger for the progress update.

boblinthorst:
Merging for testing purposes: seperate issue will be made for the updating issue still present.

PR #611, opened 2018-06-20
Add serialized content to copy and cut

atimmer:
## Summary

* [ReplacementVariableEditor] Handle cut and copy behavior in the replacement variable editor.

This change hooks into the browser copy and cut events. We then
serialize the part of editor that is currently selected if the editor is
focused. We try to put this serialized text into the clipboard data.

This means we can now copy and cut replacement variables from a
replacement variable editor.

## Test instructions

* Open the editor
* Copy a piece of text that includes replacement variables.
* Paste this into another replacement variable editor.
* You should now have the same replacement variables.
* Paste this into a text editor or the URL input.
* You should now have the text with the variables replaced with their `%%variable%%` equivalent.

Fixes https://github.com/Yoast/wordpress-seo/issues/10046

### DIFF
--- a/composites/Plugin/SnippetEditor/serialization.js
+++ b/composites/Plugin/SnippetEditor/serialization.js
@@ -1,5 +1,3 @@
-import reduce from "lodash/reduce";
-import sortBy from "lodash/sortBy";
 import { EditorState, Modifier, SelectionState, ContentState } from "draft-js";
 
 const CIRCUMFIX = "%%";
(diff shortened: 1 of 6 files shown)


composites/Plugin/SnippetEditor/serialization.js
```diff
@@ -20,48 +18,143 @@ export function serializeVariable( name ) {
...
+		} )
+		.takeUntil( function( block ) {
+			const result = lastWasEnd;
+
+			if ( block.getKey() === endKey ) {
+				lastWasEnd = true;
+			}
+
+			return result;
+		} );
```
maartenleenders: Fancy

maartenleenders:
CR 👍

PR #612, opened 2018-06-20
Implement an option for the snippet editor paper style

### LINKED ISSUE #609: Make paper around snippet editors on Search appearance settings optional
In the new design of the Search appearance settings, the snippet editors are embedded in a large collapsible header:
 
Because the panel is already white, we don't need the white paper (incl. box shadow etc) around the snippet editors anymore. Add a prop to the component that specifies whether or not the component needs a paper background.

igorschoester:
## Summary

* Implements an option for the paper style in the styled section component.

* Adds `hasPaperStyle` prop that removes the box-shadow and changes the background color to transparent.
* Removes the left and right padding in the SettingsSnippetEditorFields.

## Test instructions

* Check if changing the prop to false removes the box-shadow and changes the background color to transparent. Tip: you can use the react developer tools extension in chrome to find the react component and toggle the prop.

Fixes #609


boblinthorst:
CR: ok 👍

PR #618, opened 2018-06-25
Simplify replacement variable insert button alignment.

### LINKED ISSUE #583: Create a better solution for the breakpoint width in the SnippetEditorFields
In the PR https://github.com/Yoast/yoast-components/pull/568 a prop called `mobileWidth` is introduced. This represents a breakpoint for when the `SimulatedLabel` `TriggerReplacementVariableSuggestionsButton` are to close together.

But for example when the user has another language, the label could well be much wider than expected. Therefor we need a more robust solution here: actually checking the width needed.

### LINKED ISSUE #594: Fix alignment of snippet variable insert buttons in settings
In the Settings > Search Appearance, the snippet variable inserter button is (unwantedly) located below the input field title, instead of next to it. When you make the window narrower and wider again, the buttons are (correctly) next to the input field title:

button3
_My apologies for the bad gif quality. The filesize of a higher quality gif was too large to upload_

Find a way to make the variable insert button on the settings pages behave properly without using a timeout. See https://github.com/Yoast/yoast-components/pull/567/files/9d941e54ba280f095d9c8b1cd8721b125c919a3a..d49c1fe99cdc9ac4abfc2ea3c1d12079382bc051#diff-a6998dc55739e5697356856577694014R96
Without the time-out the button ended up behind the input field.

afercia:
## Summary

Fixes the “insert snippet variable” button alignment by using only CSS and removing any JS widths calculation.

- Flexbox
- the "labels" have a min-width of 200 pixels, this should give enough room for longer translations on narrow screens; on large screens, the labels are allowed to grow

## Test instructions

- in the standalone version you can test just the meta box snippet editor
- in order to test also the Settings pages variables editor, you need to yarn-link to the plugin branch `release/7.7` and test in the plugin
- test the button is aligned horizontally with the labels on large screens
- on small screens the button should go below the label
- note: currently there's an issue with the "Archives > Special pages" section, we'll see if there's a way to fix it

Fixes #594 
Fixes #583

### DIFF
--- a/composites/Plugin/SnippetEditor/components/ReplacementVariableEditor.js
+++ b/composites/Plugin/SnippetEditor/components/ReplacementVariableEditor.js
@@ -132,7 +132,6 @@ ReplacementVariableEditor.propTypes = {
 
 ReplacementVariableEditor.defaultProps = {
 	replacementVariables: [],
-	mobileWidth: 356,
 };
 
 export default ReplacementVariableEditor;
--- a/composites/Plugin/SnippetEditor/components/ReplacementVariableEditor.js
+++ b/composites/Plugin/SnippetEditor/components/ReplacementVariableEditor.js
@@ -71,7 +71,6 @@ class ReplacementVariableEditor extends React.Component {
 			isHovered,
 			replacementVariables,
 			recommendedReplacementVariables,
-			styleForMobile,
 			editorRef,
 			placeholder,
 		} = this.props;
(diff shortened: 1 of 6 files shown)


afercia:
OK so the "Archives > Special pages" part is a wordpress-seo specific implementation and needs to be addressed there, will push a PR. See `wordpress-seo/js/src/components/SettingsTitleReplacementVariableEditor.js`

igorschoester:
I tend to think this solves this issue also: https://github.com/Yoast/yoast-components/issues/583
Would you agree? @afercia @IreneStr

afercia:
Yes I think so 🙂

IreneStr:
Seems like it 🙂  I've added it to the PR message

maartenleenders:
Acceptance 👍

PR #638, opened 2018-07-02
Remove ref to solve console error

IreneStr:
## Summary

* Fixes a bug where console errors would be thrown because `ref` instead of `innerRef` was used in the `ReplacementVariableEditor`.


IreneStr:
Merging after discussing with @moorscode

PR #639, opened 2018-07-02
Introduce wrapper in mock to pass innerRef as a ref and re-enable tests

IreneStr:
## Summary

* N/A

* Introduced a wrapper in the mock for ReplacementVariableEditorStandalone which passes `innerRef` as a `ref`. ReplacementVariableEditorStandalone needs the ref prop because the `focus` function needs it.
* Re-enable tests that are dependent on this mock.

## Test instructions

* Run `yarn test` and check that all tests pass.


maartenleenders:
CR and Acceptance 👍

Issue #633, opened 2018-06-29
Buttons: fix the CSS media query hack for IE11


afercia:
Testing again the Buttons in IE 11, the ones based on the `Button` component look broken while the ones based on `YoastButton` look OK.

We've implemented a couple CSS hacks for the buttons to work correctly in Safari and IE11, as there are specific and known bugs related to button elements and CSS flexbox. See https://github.com/Yoast/yoast-components/pull/262 and https://github.com/Yoast/yoast-components/pull/284

Turns out the CSS media query we use to target IE11 works only if there are no spaces between the brackets. At some point, spaces where added as part of a coding standard review 😬and then removed back but just for the `YoastButton` component. We missed to remove them also for the `Button` component.

The best way to see the IE11 bug is by increasing the Button `settings.minHeight` value. Notice the wrong alignment of text and icons:

![screenshot 17](https://user-images.githubusercontent.com/1682452/42096794-a03bd3ac-7bb6-11e8-83a7-e0f4f94e5f0f.png)

afercia:
Going to push a quick, tiny PR :)

Issue #632, opened 2018-06-29
Buttons: addHoverStyle, addFocusStyle, addActiveStyle default props have no effect


afercia:
Continuing the investigation on button while attempting to build a test page for buttons, see #631 ,I've noticed that after the changes in https://github.com/Yoast/yoast-components/commit/d63570b1568902322370beb09196bd3d0985d803 the default props for the buttons hover, focus, and active styles don't have any effect.

Previously, these styles used the colors from the color palette. Now, they use colors from the props, with some default values set as `defaultProps`. However, these defaultProps are set on the wrapping styling functions: they're not component so props won't work there.

See for example https://github.com/Yoast/yoast-components/blob/11dd35f4aa9a3a23994e22294581e83240172852/composites/Plugin/Shared/components/Button.js#L99 where `props` is undefined within that function.

Instead, these default props should be set on the Button component itself, or any other component that makes use of `addHoverStyle()`, `addFocusStyle()`, and `addActiveStyle()`. See for example the `YoastButton` which makes use of props in its `addButtonStyles()` function, but the `textColor` is set to the component itself: https://github.com/Yoast/yoast-components/blob/11dd35f4aa9a3a23994e22294581e83240172852/composites/Plugin/Shared/components/YoastButton.js#L137

As a consequence, all buttons that are intended to rely on the default props, don't display any style change for the hover, focus, and active states. For example, the "Edit snippet" button is supposed to change the following styles when hovered:
- color (it works because it uses a color from `$colors`)
- background (doesn't work, it's a prop)
- border (doesn't work, it's a prop)

This is how it looks like on hover:

<img width="246" alt="now" src="https://user-images.githubusercontent.com/1682452/42091733-c05f4566-7ba6-11e8-9f1e-c75480c2231a.png">

And this is how it's supposed to look like:

<img width="260" alt="should" src="https://user-images.githubusercontent.com/1682452/42091744-c788f8e6-7ba6-11e8-9f49-62970f046b5b.png">

Note: this applies also to the `LinkButton` component, it should set the default props to `BaseLinkButton`.

Needs #649

afercia:
Moving to queue / dev because it needs to be fixed to properly implement #627

Issue #631, opened 2018-06-29
The Button component fails to render a correct font-size


afercia:
Looking at #627 I'm building a page in the standalone yoast-component app to render all the buttons, as I think we're at a point where checking their behavior and visual appearance would be beneficial. While at it, I've noticed the `Button` component font-size is not applied correctly.

The only difference between `BaseButton` and `Button` should be that the latter has a smaller font-size (0.8rem). However it seems this fails, because of the order the CSS classes are applied to the button.

Right now, all buttons have the same font size:

![screen shot 2018-06-29 at 11 24 03](https://user-images.githubusercontent.com/1682452/42084945-6a6407c6-7b8f-11e8-940a-5681db79fb77.jpg)

The CSS class with `font-size: 0.8rem;` gets applied first and then overridden by the one with `font-size: inherit;`.

afercia:
Same for the `LinkButton`, which is an `a` element styled like a button:

<img width="840" alt="screen shot 2018-06-29 at 11 57 56" src="https://user-images.githubusercontent.com/1682452/42086567-c714f59e-7b93-11e8-87ed-5b6b1bd4d12c.png">

afercia:
I have a quick PR ready for this but maybe we should discuss a bit what's really needed here. /Cc @IreneStr 

Currently, the buttons always inherit the font-size from the parent and ancestors. This may or may not be desired, but seems to work in most of the cases. For example in WordPress this would inherit a 13 pixels font-size in most of the cases.

On the other hand, I'm not sure what the original intent of `font-size: 0.8rem;` is, as this value translates to a computed font-size of 12.8 pixels. At the very least, it should be `0.8125rem`.

With a proper fix in place, the base buttons would keep a font-size of 16 pixels while all the other ones would have a font-size of 13 pixels. 

<img width="838" alt="screen shot 2018-06-29 at 16 47 25" src="https://user-images.githubusercontent.com/1682452/42099444-b8a1902e-7bbd-11e8-82d8-939a7e07c0c2.png">

However, I'm not sure this would be desired in all cases and maybe we should consider to pass the font-siz as a prop with a default value?

IreneStr:
> However, I'm not sure this would be desired in all cases and maybe we should consider to pass the font-siz as a prop with a default value?
[...]

PR #648, opened 2018-07-06
Add buttons page

### LINKED ISSUE #647: Add buttons test page
Splitting this out from #631 

While testing buttons, it seemed logical to add a page to the standalone app to test all the buttons. 

Currently, we don't have any place where to check (at least visually) the way buttons are rendered and potential layout bugs. See for example #632.

afercia:
## Summary

Adds a buttons test page to the examples standalone app

The intent of this page is to have a place where to check all the different buttons we have in yoast-components and allow to check they work correctly (at least visually). For example, it would have been useful to have this to prevent some regressions, see #631, #632, and #633.

Please check if I've missed any button type (they can always be added later if needed)

Fixes #647

### DIFF
--- a/App.js
+++ b/App.js
@@ -8,6 +8,7 @@ import Loader from "./composites/basic/Loader";
 import HelpCenterWrapper from "./app/HelpCenterWrapper";
 import SidebarCollapsibleWrapper from "./app/SidebarCollapsibleWrapper";
 import SnippetEditor from "./app/SnippetEditorExample";
+import ButtonsWrapper from "./app/ButtonsWrapper";
 
 // Required to make Material UI work with touch screens.
 import injectTapEventPlugin from "react-tap-event-plugin";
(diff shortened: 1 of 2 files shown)


boblinthorst:
Code review and acceptance is fine. I'm not sure why some of the separators are placed where they are, but as this is a page for developers, so we don't need to stress the details.

I took a quick look through yoast-components, checking if there are buttons that should be part of this page as well. I think you've got the important button types covered. 

The step button is somewhat separate, but we're not really planning on reusing that one I think.

I'm also somewhat amused by the iconbuttonToggle. As far as I can see, the state is the only thing changed by the toggling of that button, but for some reason, the toggle state seems to influence the visibility of the separators above and below the formbutton.

afercia:
>  but for some reason, the toggle state seems to influence the visibility of the separators above and below the formbutton for me.

Yep I think that's because the button has a different height when it changes state... one more thing to check. Thanks!

PR #649, opened 2018-07-09
Merge content analysis in sidebar feature branch into develop

IreneStr:
## Summary

This PR can be summarized in the following changelog entry:

* Add `setOverallSeoScore` and `setOverallReadabilityScore` reducer.
* Add help text to the content analysis in the standalone.

## Relevant technical choices:

*

## Test instructions

This PR can be tested by following these steps:

* Check if the help text is added to the content analysis in the standalone.
* Link this branch to Free. Look in the redux whether the SEO score and readability score are updated correctly.

Requires https://github.com/Yoast/wordpress-seo/pull/10305

PR #650, opened 2018-07-09
Add a component to render a button with icon and text label

### LINKED ISSUE #627: Create a sidebar button component
Create a sidebar button component that can be used for the following things: 
schermafbeelding 2018-06-28 om 09 02 23
NB. Please think about a better name than sidebarButton, if possible.

!! **Branch from feature/content-analysis-in-sidebar** !!

#### Context of the buttons:
sidebar 7 expanded

Needs #649

### LINKED ISSUE #632: Buttons: addHoverStyle, addFocusStyle, addActiveStyle default props have no effect

afercia:
## Summary

* Added a `IconLabelledButton` component to render a button with visible textual label

In order to re-use some wrapper functions used for the buttons (addHoverStyle, addFocusStyle, addActiveStyle) I had to fix #632 first. To test this part, please follow these steps:
- on the **develop** branch, go in the buttons test page
- check the first 7 buttons and verify:
  - on hover, there's no visible change other than the text color
  - on focus, there's no visible change other than the text color and the focus blue "outline"
  - on active (while clicking the button), there's no visible change other than the text color and the focus blue "outline"
- switch to this branch and verify:
  - on hover, also the background and border change 
  - on focus, also the background changes 
  - on active, the buttons look "pressed", with an inset box-shadow

With regards to the new `IconLabelledButton` button, I've added it to the Buttons test page so you can just run the standalone app (`yarn start`) and go to localhost:3333. See in the screenshot below:

screen shot 2018-07-09 at 10 23 54

A few questions:
- should the button text be bold? Unsure if the semi-bold in the screenshot in the issue is just how macOS renders text or if it's a real bold
- is there any case where we'd want the icon and text of different colors? I guess no, and if so I'd try to slightly simplify the props
- do we need more props (e.g. icon size)?
- in the issue there's no design for the hover/focus/active states, so I've implemented just a basic focus style and there are already props for the states
- any better idea for the component's name `IconLabelledButton`?

Fixes #627 
Fixes #632 

Needs #649


maartenleenders:
CR 👍

maartenleenders:
Acceptance 🚧 

See video!
Click focus (active?) does not seem to work in Safari and Firefox. Focussing via the TAB key does show the blue border. Chrome works fine.


stylingerrors

afercia:
Well, this is due to the different way browsers apply the `:focus` and `:active` states when _clicking_ a button. Testing on the mac:

While clicking a button (and keeping the mouse/trackpad pressed), Chrome considers the button "focused" and "active". Other browsers don't, and they just apply the "active" style, if any.

To confirm the different behavior, add this to the base button style:
```
&:focus:active {
	background: yellow !important
}
```

Chrome applies the yellow background, this means it considers the button focused and active at the same time. Firefox and Safari don't.

To add more fun, the behavior is also different when activating the button with the Spacebar:
- tab to the button and press Spacebar
- Chrome and Safari show the yellow background
- Firefox doesn't

Worth noting the active style can be passed via props, so it can be made the same of the focus style, if needed. By default, there's no active style.

For completeness, I guess we need to add the text color prop to the active style so all the three state can handle `color`, `background-color`, and `border-color`.

The `box-shadow` is hardcoded, I haven't touched that but it can be tweaked if need be.

afercia:
Please also notice the few questions in the first comment, most of them related to design /Cc @hedgefield

hedgefield:
- The text is fine, font weight can be based on default links.
- No different colors between text and icon are needed.
- A prop for icon size is also not needed.
- I like the custom focus indicator you made. Might be nicer if the color was blue, to match the text color. So on hover it might invert the colors basically, blue button shape and white text.

afercia:
In the last commit:
- the icon color now uses the text color prop
- added a hover style with inverted colors
- matched the active style with the hover style
- improved the order of the rendered style elements

There's still one thing to address though, due to inconsistency across browsers.

afercia:
OK so I've spent some more time debugging and found out that to minimize inconsistencies across browsers it's best to match the hover, focus, and active styles, at least for the colors.

More importantly, there's a fundamental downside in the way we're using the `addHoverStyle`, `addFocusStyle`, and `addActiveStyle` functions together with the default props. In native CSS, it would be extremely easy to get what we need. Specifically, we need the states pseudo classes printed out in a specific order:

:hover
:focus (must override :hover)
:active (must override :hover and :focus)

Instead, with styled-components and the above functions, things work differently. For example, when passing props for the `:hover` state, the styles generated via default props get printed out first and the style generated via the passed prop gets printed out later. We'll have something like this, in this order:
```
default `active` style coming from default props:
.ldDEIv:active{color:#fff;background-color:#0066cd;box-shadow:inset 0 2px 5px -3px rgba( 0,0,0,0.5 );}

`hover` style passed via props:
.gyHKFR:hover{color:#fff;background-color:#a4286a;}
```

Now, when a button is hovered and pressed, the `:hover` style will prevail and that's not what we want. I guess this is just how styled-components works, it prints the default style before the styles passed via props, which makes generally sense because custom styles must override the default ones. Unfortunately, in the case of states pseudo classes this can lead to unwanted results.

So, when passing custom props, there's the need to pass _all_ the props related to the states.

maartenleenders:
CR 👍

maartenleenders:
Acceptance 👍

PR #651, opened 2018-07-09
Changed video alignment in help center when no ads present

Dieterrr:
## Summary

* Fixes alignment of video in help center when no ads are present.

* Added a prop based on VideoContainer determining whether or not there are ads present. 

## Test instructions

* Checkout this branch.
* Run `yarn install` on this branch.
* Link this branch to WordPress SEO Premium. ( run`yarn link` in this branch, and run `yarn link yoast-components` in wordpress-seo-premium.
* Go to the help center, first tab: "Video Tutorial" and see that the video is now correctly aligned in the center. 
* Follow the same steps for WordPress SEO to make sure the alignment is still to the left and allows space for advertisements on the right side. 

Fixes #[10106](https://github.com/Yoast/wordpress-seo/issues/10106)

### DIFF
--- a/composites/HelpCenter/views/VideoTutorial.js
+++ b/composites/HelpCenter/views/VideoTutorial.js
@@ -15,7 +15,7 @@ const VideoTutorialContainer = styled.div`
 `;
 
 const VideoContainer = styled.div`
-	float: left;
+	${ props => props.hasAdvertisements ? "float: left;" : "margin: auto;" }
 	width: ${ VIDEO_WIDTH };
 
 	@media screen and ( max-width: ${ breakpoints.tablet } ) {
@@ -24,6 +24,10 @@ const VideoContainer = styled.div`
 		margin: 0 auto;
 	}
 `;
+
+VideoContainer.propTypes = {
+	hasAdvertisements: PropTypes.bool.isRequired,
+};
 
 const VideoDescriptions = styled.div`
 	margin-left: ${ VIDEO_WIDTH };
@@ -102,10 +106,14 @@ VideoDescriptionItem.defaultProps = {
 export default function VideoTutorial( props ) {
 	return (
 		<VideoTutorialContainer className={ `${ props.className }__container` }>
-			<VideoContainer className={ `${ props.className }__video-container` }>
+			<VideoContainer
+				className={ `${ props.className }__video-container` }
+				hasAdvertisements={ props.paragraphs.length > 0 }
+			>
 				<YouTubeVideo
 					src={ props.src }
-					title={ props.title } />
+					title={ props.title }
+				/>
 			</VideoContainer>
 			<VideoDescriptions className={ `${ props.className }__descriptions` }>
 				{ props.paragraphs.map( paragraph => {
--- a/composites/HelpCenter/views/VideoTutorial.js
+++ b/composites/HelpCenter/views/VideoTutorial.js
@@ -104,18 +104,20 @@ VideoDescriptionItem.defaultProps = {
  * @returns {ReactElement} The VideoTutorial component.
  */
 export default function VideoTutorial( props ) {
+	const hasParagraphs = props.paragraphs.length > 0;
+
 	return (
 		<VideoTutorialContainer className={ `${ props.className }__container` }>
 			<VideoContainer
 				className={ `${ props.className }__video-container` }
-				hasParagraphs={ props.paragraphs.length > 0 }
+				hasParagraphs={ hasParagraphs }
 			>
 				<YouTubeVideo
 					src={ props.src }
 					title={ props.title }
 				/>
 			</VideoContainer>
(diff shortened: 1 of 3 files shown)


composites/HelpCenter/views/VideoTutorial.js
```diff
@@ -102,10 +106,14 @@ VideoDescriptionItem.defaultProps = {
 export default function VideoTutorial( props ) {
 	return (
 		<VideoTutorialContainer className={ `${ props.className }__container` }>
-			<VideoContainer className={ `${ props.className }__video-container` }>
+			<VideoContainer
+				className={ `${ props.className }__video-container` }
+				hasAdvertisements={ props.paragraphs.length > 0 }
```
andizer: I would change the naming. You want to align the video if there is no paragraph. Advertisements is more like an implementation details. Can you change it to: hasParagraphs or something like that.

Dieterrr: 👍

andizer:
CR 2 Done 👍

afercia:
Acceptance 🚧 

Worth noting that, when no paragraphs are passed, the paragraphs container is still rendered. It's empty, but still there and with a 0 pixels height and a 560 pixels left margin. Even if _at the moment_ it doesn't seem to harm anything, any future variation in the layout might produce unexpected results. In the screenshot below, I've made it visible for testing purposes:

screen shot 2018-07-11 at 12 46 26

I'd suggest to render it conditionally only if there are paragraphs. For clarity, something like 
`const hasParagraphs = props.paragraphs.length > 0;`

then reuse this for the VideoContainer prop and also wrap the VideoDescriptions in a simple evaluation with a logical AND:
`{ hasParagraphs && <VideoDescriptions ...`

afercia:
In the last commits:
- avoided to render the empty paragraphs (video descriptions a.k.a. "ads") container
- at this point the VideoTutorial `paragraphs` shouldn't be required and by default should be an empty array, to not rely on the implementation in the plugin
- added a snapshot test for the VideoTutorial with no paragraphs

I'd consider to add a VideoTutorial with no paragraphs in the standalone app examples.

composites/HelpCenter/views/VideoTutorial.js
```diff
@@ -15,7 +15,7 @@ const VideoTutorialContainer = styled.div`
 `;
 
 const VideoContainer = styled.div`
-	float: left;
+	${ props => props.hasParagraphs ? "float: left;" : "margin: auto;" }
```
andizer: `margin: auto` is also taking the top and bottom margin, is that intended? I think `margin: 0 auto` would be a better solution. Perhaps `display:block` might be needed as well, because of cross-browser.

composites/HelpCenter/views/VideoTutorial.js
```diff
@@ -100,22 +104,28 @@ VideoDescriptionItem.defaultProps = {
...
+				hasParagraphs={ hasParagraphs }
+			>
 				<YouTubeVideo
 					src={ props.src }
-					title={ props.title } />
+					title={ props.title }
+				/>
 			</VideoContainer>
-			<VideoDescriptions className={ `${ props.className }__descriptions` }>
+			{ hasParagraphs && <VideoDescriptions className={ `${ props.className }__descriptions` }>
```
andizer: This is hard to read, when hasPararaph is true, the component will be rendered I guess? It might be a better idea to wrap this in a function: `renderVideoDescriptions`. In that function you can do an early `return null` and returns the rendered videodescriptions as default like:

```js
function renderVideoDescriptions( props ) {
     if(  props.paragraphs.length > 0 ) {
        return null;
     }

     // Render the component.
     ....
} 
```

afercia: Hm I tend to disagree. This is pretty standard in Gutenberg, and personally I think it's an elegant and concise way to render conditionally. Why we should return null?

boblinthorst:
I've addressed the first point in the review. 

The second point I decided not to change for now. I can see how it is less readable than an if-statement, which signals clearly to the reader that conditional code is coming up. But it is also a structure already used in this project on multiple places. And it's more concise than the alternative.

https://github.com/Yoast/yoast-components/blob/5af57eecab39bee08288d30ab0f87ab09adccfa8/composites/Plugin/Shared/components/Collapsible.js#L230

https://github.com/Yoast/yoast-components/blob/5af57eecab39bee08288d30ab0f87ab09adccfa8/composites/Plugin/Shared/components/YoastModal.js#L54-L55

https://github.com/Yoast/yoast-components/blob/5af57eecab39bee08288d30ab0f87ab09adccfa8/composites/basic/Notification.js#L87-L92

andizer:
Missing some essential steps for testing this like `yarn install` after linking the yoast-components and `grunt build:js build:css` to build the assets.

andizer:
CR & Acceptance done.

PR #664, opened 2018-07-13
Add ThemeProvider to pass a language direction prop to the editor example

### LINKED ISSUE #663: Snippet editor in the standalone app crashes
Recently, a `theme` was added to the snippet editor in order to detect and support RTL languages and pass a corresponsing prop to the replacement variables editor (draft.js).

This works in the plugin but the standalone app on localhost:3333 doesn't use a theme. Thus, when clicking on the "Edit snippet" button, the app shows the error boundary alternative content:

`Uncaught TypeError: Cannot read property 'isRtl' of undefined`

screen shot 2018-07-13 at 16 12 21

RIght now, testing on the standalone app is impossible. Passing a `theme` prop would be trivial, but I guess for proper testing we should try to mimic as much as possible the plugin implementation, using a `ThemeProvider` wrapper.

afercia:
## Summary

* Fixes an error in the snippet editor standalone example

- uses ThemeProvider to pass a theme prop with a proper `isRtl` setting
- `isRtl` is handled in the state
- adds a button to switch language direction in the editor, thought this was a nice-to-have for testing

Notes:
- the language switch is not immediately applied to the replacement variable editors, not sure why; there's the need to click on them or click the preview elements to trigger the re-rendering
- consider this switches the language just on the draft.js editors and for components that use the state `isRtl` property; for a proper, full, language direction switch many other things are needed for example setting a `dir` attribute on the document
- for the future: maybe consider to extend the language switch to all the examples, I guess the components will support more and more RTL languages and this would help in testing

## Test instructions

- yarn start and go to localhost:3333
- in the snippet preview tab, click on the "Edit snippet" button and verify the editor opens without errors
- click the "Change Editor language direction" then click the editors and check the text directionality changes
- note: a few other elements in the page are affected by the language directionality switch, some of them (button icons) should be addressed separately

Fixes #663

### DIFF
--- a/app/SnippetEditorExample.js
+++ b/app/SnippetEditorExample.js
@@ -1,6 +1,6 @@
 // External dependencies.
 import React, { Component } from "react";
-import styled from "styled-components";
+import styled, { ThemeProvider } from "styled-components";
 import debounce from "lodash/debounce";
 import MetaDescriptionLengthAssessment from "yoastseo/js/assessments/seo/metaDescriptionLengthAssessment";
 
@@ -11,6 +11,10 @@ const Container = styled.div`
 	background-color: white;
 	margin: 5em auto 0;
 	padding: 0 0 10px;
+`;
+
+const LanguageDirectionContainer = styled.div`
+	text-align: center;
 `;
 
 const replacementVariables = [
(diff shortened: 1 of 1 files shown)


maartenleenders:
CR 👍

maartenleenders:
Acceptance 👍 
Not perfect yet, but split out in: https://github.com/Yoast/yoast-components/issues/665

PR #672, opened 2018-07-25
Expose makeOutboundLink for UpsellBox in wpseo

### LINKED ISSUE #661: Create upsell component
## Context
schermafbeelding_2018-07-13_om_09_15_17

Create an upsell component that is going to be used in the sidebar and metabox. The component should contain one or more paragraphs, one or more bullet points (with plusses), and a button with a paragraph underneath. The paragraph under the button and the button should be 'connected' to each other using `aria-describedby`.

maartenleenders:
## Summary

* We now expose makeOutboundLink, inside a utils object.
* Added `Clipboard` and `a11y-speak` to the `package.json` because tests were failing.

* We needed makeOutboundLink() in wordpress-seo, but were asked by @atimmer to not use paths when importing. Thus, we needed to expose makeOutboundLink in a `utils/index.js` file.

## Test instructions

* Checkout this branch and verify that all tests succeed, also when linking to wordpress-seo.

Fixes #661

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,6 @@
   "author": "Yoast",
   "license": "GPL-3.0",
   "dependencies": {
-    "a11y-speak": "git+https://github.com/Yoast/a11y-speak.git#master",
     "@wordpress/a11y": "^1.0.7",
     "@wordpress/i18n": "^1.1.0",
     "algoliasearch": "^3.22.3",
--- a/package.json
+++ b/package.json
@@ -43,9 +43,11 @@
   "author": "Yoast",
   "license": "GPL-3.0",
   "dependencies": {
+    "a11y-speak": "git+https://github.com/Yoast/a11y-speak.git#master",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -43,9 +43,11 @@
   "author": "Yoast",
   "license": "GPL-3.0",
   "dependencies": {
+    "a11y-speak": "git+https://github.com/Yoast/a11y-speak.git#master",
     "@wordpress/a11y": "^1.0.7",
     "@wordpress/i18n": "^1.1.0",
     "algoliasearch": "^3.22.3",
+    "clipboard": "^1.5.15",
```
atimmer: This should not have any effect. The package on npm specifies `ClipboardJS` as the API. We never use that. We use `Clipboard` which is a browser API.

maartenleenders: We had some issues with running `yarn test` that revolved around clipboard though. We may have misinterpreted the error though. Can you advise on what we should do instead?

Issue #671, opened 2018-07-25
Styling improvements for the Analysis results. 


boblinthorst:
Split off from https://github.com/Yoast/yoast-components/pull/492.

The PR became way too big, and addressing the styling comments in the PR would make it even bigger than it already is.

In the words of @maartenleenders:
The styling in the ContentAnalysis seems to be slightly below our standards, but that is my opinion. I think we'd do well to give everything a little more 'breathing room' with padding/margins. @hedgefield has offered to sit in on a little styling session to improve things. I would also be fine if you split this up into a new issue, as this PR has become quite big.

hedgefield:
I checked out the components standalone app and did some tweaking. Here are some improvements:

- For each collapsible section, add `margin-bottom: 8px`
- For each collapsible section header, set `font-size: 1.1em` and `padding: 8px 16px`
- For each analysis result, set `padding: 8px 16px`

I assume the margin/padding around all the components is set globally on the sidebar element? Is it possible to remove it there and add it to each component individually? It's a little cumbersome, but it produces better results for the focus indicator. Or would that be too much inefficient CSS?

![3](https://user-images.githubusercontent.com/6073772/43311117-bcb113c4-9189-11e8-9ee8-c70ea8d99c24.png)

afercia:
One important thing to consider is that the ContentAnalysis uses a custom implementation of the "collapsible header", duplicating great part of the code. I'd tend to think it shouldn't do this, and just reuse the `Collapsible` component which is going to be refactored in #653 and #654

afercia:
> For each collapsible section header, set font-size: 1.1em

I'd need the value in pixels for this, as the em one varies depending on the context (standalone app or within the meta box) and i've also made other changes.

> For each analysis result, set padding: 8px 16px

Do you mean each group of results or each single item in a group of results?

hedgefield:
> I'd need the value in pixels for this, as the em one varies depending on the context (standalone app or within the meta box) and i've also made other changes.

I'd say 14 then, just a tad bigger than the default 13px sidebar font size.

> Do you mean each group of results or each single item in a group of results?

Each single item.

afercia:
> Each single item.

OK but I'd suggest to check it again once it's in the meta box. Right now, with the changes made on #676 the items already have a good spacing (see below). I'm afraid adding more padding will make them way too spaces, but again this needs to be checked in the meta box.

[...]

PR #676, opened 2018-07-30
653 654 collapsible header icon keyword

### LINKED ISSUE #653: Create a collapsible header with keyword and icon
Create a collapsible header with keyword and icon. This header will be used both in the sidebar, as well as in the metabox. This means its width should be flexible. We already have a collapsible header with icon but without keyword (https://github.com/Yoast/yoast-components/issues/653), so please use that one as a basis. The color of the icon and the icon itself should be able to be changed based on passed props.

NB. In the designs, there is a keyword color difference (grey vs black) between the metabox and sidebar. Please implement it in black. 

## Sidebar
schermafbeelding_2018-07-10_om_09_03_11

## Metabox
schermafbeelding_2018-07-10_om_09_05_01

Needs https://github.com/Yoast/yoast-components/issues/653

### LINKED ISSUE #654: Create a collapsible header with icon
Create a collapsible header with icon. This header will be used both in the sidebar, as well as in the metabox. This means its width should be flexible. We already have a collapsible header without icon, so please use that one as a basis. The color of the icon and the icon itself should be able to be changed based on passed props.

## Sidebar
schermafbeelding_2018-07-10_om_11_45_47

## Metabox
schermafbeelding_2018-07-10_om_11_45_53

## SVG icons
[Zip with smiley SVGs](https://github.com/Yoast/yoast-components/files/2180173/score.smileys.zip). The grey icon is a bullet without a face.
Colors:
#008800 (green)
#f5c819 (yellow)
#dc3232 (red)
#a0a5aa (grey)

afercia:
## Summary

* Adds a subtitle to the collapsible header component

- there was the need to make changes to the `Collapsible` component, consequently the `ContentAnalysis` needed some changes too
- the new sub title is passed via the `subTitle` prop
- removes `AnalysisCollapsible` (this should go in the changelog) and use `Collapsible` directly: `AnalysisCollapsible` was duplicating 95% of the code from `Collapsible`
- adds new props to control the SVG icons, including their `ViewBox`: this is necessary because some icons have been coded with a different ViewBox size, thus affecting the way they're rendered
- makes it possible to control the left and right icons independently
- refactors `ContentAnalysis` and uses a `renderCollapsible()` helper function to simplify code
- various CSS adjustments

Ideally, the `Collapsible` header component should be as style-agnostic as possible. For the header itself, it's difficult to style it independently other than passing props. Instead, the content of the collapsible should be styled independently using a specific styled component.

Partially fixes also #671 which will need some adjustments in the plugin. As requested, the left and right paddings should be set on the component itself. This means the StyledSection used in the plugin should be adjusted consequently. Screenshot of the analysis in the plugin (notice the left and right paddings to reset):

screen shot 2018-07-29 at 17 21 39

## Test instructions

- check the standalone app (yarn start and go to localhost:3333) and then go in the "Collapsibles" tab
- verify all the collapsible examples there work as expected
- note: in this page I've added a few styles to emulate the WordPress font metrics
- note about the focus style: difficult because there's the need to override some styles that come from the button; I've made it in a way that there's also a subtile feedback when clicking the collapsible
- note about the arrow icons on the right: difficult to align them vertically when there's a subtitle, ideally they should be shifted a few pixels to the bottom
- go in the Content Analysis tab
- verify everything works correctly
- the Content Analysis collapsibles already implement some of the CSS adjustments requested in #671

Fixes #653 
Fixes #654


composites/Plugin/Shared/components/Button.js
```diff
@@ -268,15 +268,23 @@ IconButton.defaultProps = {
  * @returns {ReactElement} Styled icon button.
  */
 export const IconsButton = ( props ) => {
-	const { children: text, prefixIcon, prefixIconColor, suffixIcon, suffixIconColor } = props;
-
-	const newProps = omit( props, [ "prefixIcon", "prefixIconColor", "suffixIcon", "suffixIconColor" ] );
+	const {
```
Xyfi: I don't think we want to pass all these props to the button?
I suggest using object destructuring.
```
const {
    extractedProp1,
    extractedProp2,
    ...buttonProps
} = props;
```

composites/Plugin/Shared/components/Button.js
```diff
@@ -268,15 +268,23 @@ IconButton.defaultProps = {
...
+		children: text,
```
Xyfi: You rename the children prop explicitely to text, but the `propTypes` of the component suggest it can also be a element or multiple elements.

composites/Plugin/Shared/components/Button.js
Xyfi: Also, what a godawful amount of props. Consider grouping prefix and suffix props, so you get `props.suffix.icon`, `props.suffix.iconColor`, and so on. I think this improved readability overall.

style-guide/_colors.scss
```diff
@@ -22,13 +22,14 @@ $palette_blue_dark:          #084a67; // Safe to use with $palette_white – $pa
 $palette_green:              #77b227; // Safe to use with $palette_black.
 $palette_green_light:        #7ad03a; // Safe to use with $palette_black.
 $palette_green_medium_light: #64a60a; // Safe to use with $palette_black or $palette_white bold > 18.5px.
-$palette_green_medium:       #008a00; // Safe to use with $palette_white and $palette_black.
+$palette_green_medium:       #008a00; // Safe to use with $palette_white and $palette_black. Also used for the smiley score icon.
```
Xyfi: Shouldn't we update the `color_ok` instead?

composites/Plugin/Shared/components/Collapsible.js
```diff
@@ -11,21 +11,35 @@ const StyledContainer = styled.div`
...
 
-const StyledContent = styled.div`
-	padding: 0 15px;
-`;
-
 const StyledIconsButton = styled( IconsButton )`
 	width: 100%;
 	background-color: ${ colors.$color_white };
-	padding: 15px;
+	padding: ${ props => props.headingPadding };
```
Xyfi: I think we should refrain from creating specific props for styling, can't we use a `styled-components` way to make sure the `StyledIconsButton` has the correct padding for the components that need different paddings.

composites/Plugin/Shared/components/Button.js
```diff
@@ -268,15 +268,23 @@ IconButton.defaultProps = {
...
+		prefixIcon,
+		prefixIconColor,
+		suffixIcon,
+		suffixIconColor,
+		prefixIconViewBox,
```
Xyfi: This shouldn't exist, can't we update the SVG icon to have a `1720x1720` viewbox as well? All these `iconViewbox` props are really cluttering the code. If it's not possible to change the viewbox of the icon, we should update the viewBox of the icons that require a different view box in the `SvgIcon` component.

composites/Plugin/Shared/components/Button.js
afercia: No I haven't touched this part, it was already `children: text`

style-guide/_colors.scss
afercia: color_good, color_ok: yes we can update them but worth noting they're used also in other places, for example in the content analysis and for the progress bar. If we want to update there too, that's fine I guess.

afercia:
Addressed the main CR points except the color names, which I'd prefer to keep as they're for now. We can revisit when we're sure these colors should be updated globally.

afercia:
Note: as soon as this is merged, we should move #671 to the queue to adjust styling in the meta box.

composites/Plugin/Shared/components/Collapsible.js
```diff
@@ -152,23 +144,33 @@ CollapsibleStateless.propTypes = {
 	Heading: PropTypes.func,
 	isOpen: PropTypes.bool.isRequired,
 	onToggle: PropTypes.func.isRequired,
-	prefixIcon: PropTypes.string,
-	prefixIconCollapsed: PropTypes.string,
-	prefixIconColor: PropTypes.string,
-	prefixIconSize: PropTypes.string,
+	prefixIcon: PropTypes.shape( {
```
boblinthorst: Perhaps declare the shape seperately, and reuse this declaration wherever it's used?

Done in snippetEditorFields's descriptionLengthProgress, for instance. 

Not important enough that it has to be addressed now though.

boblinthorst:
CR: ok 👍

Xyfi:
Acceptance :+1:

Issue #660, opened 2018-07-11
Create a MultiKeywordAnalysisContainer


IreneStr:
## Context
Create a MultiKeywordAnalysisContainer which handles which keyword analysis sections are shown. This container will only be used in Premium.

<img width="519" alt="schermafbeelding_2018-07-11_om_15_30_33" src="https://user-images.githubusercontent.com/17744553/42574761-7e89753a-851f-11e8-876c-be6e406ae60d.png">

To zoom in: https://user-images.githubusercontent.com/6073772/42497707-76306f10-842a-11e8-8819-e6f4f5a84b29.png

## Glossary
1. `Focus keyword section`: this includes a focus keyword field, a synonyms field, and an analysis. When the keyword field is empty (and this component is still shown, see `Removing keywords`), the analysis is still shown (but with fewer bullets, encouraging the user to add a focus keyword, similar to the current situation in the plugin)
<img width="1284" alt="schermafbeelding 2018-07-11 om 15 38 01" src="https://user-images.githubusercontent.com/17744553/42575328-7c77cfd4-8520-11e8-842d-673763416899.png">

2. `Additional focus keyword section`: includes only a focus keyword field, no synonyms field and no analysis.
<img width="1283" alt="schermafbeelding 2018-07-11 om 15 37 47" src="https://user-images.githubusercontent.com/17744553/42575325-79dfc542-8520-11e8-9872-2c43b2d35ebb.png">

### Default
When there is not any keyword set yet, it should show a `focus keyword section`, and a `additional focus keyword section`.

### Adding keywords
When the user start typing a focus keyword in the `additional focus keyword section`, a new `additional focus keyword section` should be appended, unless that would mean that the total number of sections is 6. In that case, no section needs to be added. 5 is always the maximum number of sections/keywords.

### Removing keywords
1. When the user clears the keyword field from one of the `focus keyword section`s, and there are also no synonyms in the synonym fields. On unfocus of the keyword input field:
    - The `focus keyword section` the user has cleared should be removed.
    - If there were 5 `focus keyword section`s before (which means now there are 4), an `additional focus keyword section` should be appended at the end. 
   - There should always at least 2 sections: see Default above.

2. When the user clears the keyword field from one of the `focus keyword section`s, and there are still synonyms in the synonym fields 
    - The `focus keyword section` should stay. There should be a warning that the user should add a focus keyword.

[...]

PR #677, opened 2018-07-31
Extract yoast input field

maartenleenders:
## Summary

* SynonymsInput field now looks similar to the `KeywordInput`, and can have an explaniontext below the label.

* Needed to reuse the KeywordInput for the SynonymsInput, but found it was too specific. So we extracted some stuff and created a new SynonymsInput from the extracted components. This does not affect the existing inputs.

## Test instructions

* Checkout branch, and run `yarn start`.
* Verify in the examples (under "Keyword") that the new SynonymsInput functions and looks as expected: it should have a label, a text below the label, and a functioning input field.

Connects https://github.com/Yoast/yoast-components/issues/660


Dieterrr:
CR & Acceptance done 👍

PR #678, opened 2018-08-01
Implement cross with onClick functionality inside input container

maartenleenders:
## Summary

* Keyword input fields now have a button that allows removal of that keyword section.

* Styling the input field and overriding Gutenberg styles was very cumbersome (increasing specificity with `&&&` did not work), so for now we have used `! important`. @moorscode is doing a big styling pass, perhaps this can be fixed once he found a decent solution.

## Test instructions

* Verify in the yoast-components examples (`localhost:3333`), under keyword, that the KeywordInput now has a cross. If you click it, the console should log "CLOSED", as that is the onClick functionality we added to the example.


composites/Plugin/Shared/components/Button.js
```diff
@@ -328,3 +328,41 @@ IconsButton.defaultProps = {
...
+	border: none;
+	box-shadow: none;
+	background: none;
+`;
+
+/**
+ * A button that can be put inside a div with a borderless input-field.
+ *
+ * @param {Object}   props         The props for the button.
+ * @param {function} props.onClick The onClick function for the button.
```
jcomack: `function` should be capitalized.

composites/Plugin/Shared/components/Button.js
```diff
@@ -328,3 +328,41 @@ IconsButton.defaultProps = {
...
+ * @param {string}   props.color   The color for the icon.
+ *
+ * @returns {JSX.Element} A borderless button to be included with inputfields.
```
jcomack: I believe this should be `ReactElement` and not `JSX.Element`.

composites/Plugin/Shared/components/KeywordInput.js
```diff
@@ -112,14 +115,20 @@ class KeywordInput extends React.Component {
...
+					<KeywordField
+						aria-label={ showLabel ? null : label }
+						type="text"
+						id={ id }
+						className={ showErrorMessage ? "hasError" : null }
+						onChange={ this.handleChange }
+						value={ keyword }
+					/>
+					<YoastInputClearButton
+						onClick={ this.props.onRemoveKeyword }
```
jcomack: Why isn't this being assigned the same way as the other props are? See first line of this method.

PR #682, opened 2018-08-01
Styling improvements

### LINKED ISSUE #671: Styling improvements for the Analysis results.

moorscode:
## Summary

* _N/A_

* Added `!important` to `h2` overrides in the Collapsible, otherwise the styling of the `#postbox h2` in WordPress will not be changed - these can be seen in the components in the metabox.
* Added `hasPadding` to the Collapsible, to make the content of the collapsible to have padding and the internal items have no padding.
* Added `hasSeparator` to the Collapsible to make borders around the entries to have them visually separated

## Test instructions

* Use the PR (https://github.com/Yoast/wordpress-seo/pull/10483) on Yoast SEO to see the changes.

Fixes #671

### DIFF
--- a/composites/Plugin/Shared/components/Collapsible.js
+++ b/composites/Plugin/Shared/components/Collapsible.js
@@ -7,8 +7,19 @@ import colors from "../../../../style-guide/colors.json";
 import { IconsButton } from "../../Shared/components/Button";
 import ScreenReaderText from "../../../../a11y/ScreenReaderText";
 
+const Content = styled.div`
+	padding: 0 16px;
+	margin-bottom: 16px;
+`;
+
 const StyledContainer = styled.div`
 	background-color: ${ colors.$color_white };
+`;
+
+const StyledContainerTopLevel = styled( StyledContainer )`
+	border-top: 1px solid ${ colors.$palette_grey_light };
+	border-bottom: 1px solid ${ colors.$palette_grey_light };
+	margin-top: -1px;
 `;
 
 export const StyledIconsButton = styled( IconsButton )`
@@ -75,7 +86,8 @@ const StyledSubTitle = styled.span`
 export function wrapInHeading( Component, headingLevel ) {
 	const Heading = `h${ headingLevel }`;
 	const StyledHeading = styled( Heading )`
-		margin: 0;
+		margin: 0 !important;
+		padding: 0 !important;
 		font-weight: normal;
 	`;
 
@@ -98,6 +110,8 @@ const StyledHeading = wrapInHeading( StyledIconsButton, 2 );
  * @param {string}      props.className             The name of the collapsible CSS class.
  * @param {IconsButton} props.Heading               Heading button. May be wrapped or styled or both.
  * @param {boolean}     props.isOpen                True displays the children. False means collapsed.
+ * @param {boolean}     props.hasPadding            True adds padding to the content. False means no padding.
+ * @param {boolean}     props.hasSeparator          True displays borders around the section. False means no borders.
  * @param {function}    props.onToggle              Function to handle the Heading click event.
  * @param {object}      props.prefixIcon            Heading icon before the title.
  * @param {object}      props.prefixIconCollapsed   Prefix icon when in collapsed state.
@@ -110,8 +124,14 @@ const StyledHeading = wrapInHeading( StyledIconsButton, 2 );
  * @returns {ReactElement} A collapsible panel.
  */
 export const CollapsibleStateless = ( props ) => {
+	let children = null;
+	if ( props.isOpen ) {
+		children = ( props.hasPadding ) ? <Content>{props.children}</Content> : props.children;
+	}
+	const Container = ( props.hasSeparator ) ? StyledContainerTopLevel : StyledContainer;
+
 	return (
-		<StyledContainer
+		<Container
 			// Pass the classname to allow re-styling with styled-components.
 			className={ props.className }
 		>
@@ -130,8 +150,8 @@ export const CollapsibleStateless = ( props ) => {
 					{ props.subTitle && <StyledSubTitle>{ props.subTitle }</StyledSubTitle> }
 				</StyledTitleContainer>
 			</props.Heading>
-			{ props.isOpen && props.children }
-		</StyledContainer>
+			{ children }
+		</Container>
 	);
 };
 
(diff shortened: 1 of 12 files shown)


composites/Plugin/Shared/components/Collapsible.js
```diff
@@ -67,16 +77,21 @@ const StyledSubTitle = styled.span`
...
-export function wrapInHeading( Component, headingLevel ) {
-	const Heading = `h${ headingLevel }`;
+export function wrapInHeading( Component, props ) {
+	const Heading = `h${ props.level }`;
 	const StyledHeading = styled( Heading )`
-		margin: 0;
-		font-weight: normal;
+		margin: 0 !important;
+		padding: 0 !important;
+		font-size: ${ props.fontSize } !important;
```
jcomack: Is there a specific reason as to why `fontSize` and `fontWeight` aren't in the prop types list? Additionally, adding a default would allow you to leave out some stuff such at the `fontWeight: "normal"` on line 106.

composites/Plugin/Shared/components/Collapsible.js
```diff
@@ -67,16 +77,21 @@ const StyledSubTitle = styled.span`
 /**
  * Wraps a component in a heading element with a defined heading level.
  *
- * @param {ReactElement} Component    The component to wrap.
- * @param {int}          headingLevel The heading level.
+ * @param {ReactElement} Component        The component to wrap.
+ * @param {object}       props            The heading props.
```
jcomack: `object` should be `Object`.

composites/Plugin/Shared/components/Collapsible.js
```diff
@@ -67,16 +77,21 @@ const StyledSubTitle = styled.span`
...
+ * @param {boolean}      props.level      The heading level.
+ * @param {boolean}      props.fontSize   The heading font-size.
+ * @param {boolean}      props.fontWeight The heading font-weigth.
```
jcomack: Typo in the word `weight`.

composites/Plugin/Shared/components/Collapsible.js
```diff
@@ -98,6 +113,8 @@ const StyledHeading = wrapInHeading( StyledIconsButton, 2 );
  * @param {string}      props.className             The name of the collapsible CSS class.
  * @param {IconsButton} props.Heading               Heading button. May be wrapped or styled or both.
  * @param {boolean}     props.isOpen                True displays the children. False means collapsed.
+ * @param {boolean}     props.hasPadding            True adds padding to the content. False means no padding.
+ * @param {boolean}     props.hasSeparator          True displays borders around the section. False means no borders.
  * @param {function}    props.onToggle              Function to handle the Heading click event.
  * @param {object}      props.prefixIcon            Heading icon before the title.
```
jcomack: Although this wasn't touched: `object` should be `Object`.

composites/Plugin/Shared/components/Collapsible.js
```diff
@@ -110,8 +127,14 @@ const StyledHeading = wrapInHeading( StyledIconsButton, 2 );
  * @returns {ReactElement} A collapsible panel.
  */
 export const CollapsibleStateless = ( props ) => {
+	let children = null;
+	if ( props.isOpen ) {
+		children = ( props.hasPadding ) ? <Content>{props.children}</Content> : props.children;
```
jcomack: I believe the code-style enforces spaces after `{` and before `}`.

composites/Plugin/Shared/components/Collapsible.js
```diff
@@ -182,7 +207,7 @@ export class Collapsible extends React.Component {
 	 *
 	 * @param {object}  props                       The properties for the component.
 	 * @param {string}  props.className             The name of the collapsible CSS class.
-	 * @param {number}  props.headingLevel          Heading level: 1 for h1, 2 for h2, etc.
+	 * @param {object}  props.headingProps          Props to use in the Heading.
```
jcomack: `object` should be `Object`.

composites/Plugin/Shared/components/Collapsible.js
moorscode: They are in the props type list but as sub-item: headingProps.
As this function is not part of the class, the props are not the same props, but can be given as object constructed from anywhere.
I wanted to keep them similar to the ones used in `headingProps` to maintain reference.

jcomack:
CR 2 done 👍

jcomack:
Acceptance done 👍 

Just waiting for Travis to pass before merging.

composites/Plugin/Shared/components/Collapsible.js
afercia: We should always avoid to use `!important`.

composites/Plugin/SnippetEditor/components/ModeSwitcher.js
```diff
@@ -54,8 +53,6 @@ const DesktopButton = SwitcherButton.extend`
 const Switcher = styled.div`
 	display: inline-block;
 	margin-top: 10px;
-	margin-right: ${ getRtlStyle( "0px", "20px" ) };
-	margin-left: ${ getRtlStyle( "20px", "4px" ) };
 	border: 1px solid #dbdbdb;
```
afercia: The 4px margin was needed for RTL, though a better fix would be reversing the Edit Snippet button margin.

composites/Plugin/Shared/components/Collapsible.js
afercia: To expand on this, there's a better way to increase specificity when necessary, see 
https://github.com/Yoast/yoast-components/blob/dac4110ea3519488d0ea42e18bfbcc1e79a3b0a3/composites/AlgoliaSearch/SearchBar.js#L36

the reason why this approach is preferable is that `!important` can't be overridden other than using another `!important` while the second approach can be overridden in an easier way, if need be. Will open an issue for consideration.

Issue #665, opened 2018-07-16
Add a real switch to RTL languages in the standalone app examples


afercia:
#664 is a first step to introduce a way to test the standalone app examples with RTL languages. For now, it just switches the `isRtl` state property that is then passed to the styled-components `theme`

This is enough to switch the replacement variable editors to RTL (Draft.js has a custom way to support language direction) but it's not a real switch to RTL.

At the very least, we should find a way to set a `dir="rtl"` attribute on the `<html>` element in the `index.html` file. Ideally, also the language used for all the text in the components should be a RTL language.

Currently, a few other components react to the RTL switch because of `getRtlStyle()` but they don't render properly because of the missing `dir="rtl"` attribute. For example:
ModeSwitcher
TriggerReplacementVariableSuggestionsButton
HelpTextWrapper
EditSnippetButton
the "caret" arrows 
StyledHeading

afercia:
Switching the dir="rtl" attribute is added in #685 

To properly render in RTL, what is left here is:
-  the language used for all the text in the components should be a RTL language
- change the `lang` attribute on the `<html>` accordingly

Of course this is a bit challenging but I think it's worth considering we do need a test environment to test in RTL.

PR #685, opened 2018-08-02
Toggle - cornerstone toggle improvements

### LINKED ISSUE #681: Toggle - CornerstoneToggle improvements
Noticed there's room for a few improvements to the Toggle / CornerstoneToggle component:

functional:
- when the toggle is toggled pressing the spacebar, there's the need to prevent the default behavior otherwise the page will scroll

a11y:
- since the toggle emulates a checkbox but it's not a native `<checkbox>` element, it can't use a `<label>` element. Worth noting `<div>` elements are not [labelable elements](https://www.w3.org/TR/html5/sec-forms.html#labelable-element)
- the visible text should be just text, referenced by `aria-labelledby` on the element with role=checkbox
- `aria-label` should be removed, as we want the toggle to always have a visible text

misc:
- the visible text shouldn't have a period at the end
- the `id` is hardcoded, so just one `CornerstoneToggle` can be used in a page

afercia:
## Summary

* Toggle component accessibility and RTL support improvements

- adds a switch to RTL to all the demo pages in the standalone app
- updates the `dir` attribute on the `<html>` element when switching to RTL, this partially addresses #665
- adds Toggle and CornerstoneToggle demos to the "UI Controls" tab (formerly "Checkbox" tab) in the demo pages
- Toggle: removes the `<label>` element and uses `aria-labelledby` instead
- Toggle: passes the ID as a prop with a default value
- CornerstoneToggle: changes the text to "Mark as cornerstone content" (addresses design feedback)
- Toggle: uses `getRtlStyle()` for RTL languages and adjusts various CSS 
- Toggle: prevents the page from scrolling when using the Spacebar 
- Toggle: when using the Spacebar, switch the toggle on `keyup` instead of `keydown`, as that's the native behavior
- Toggle: makes `onToggleDisabled()` work when using the Spacebar 
- Toggle: uses `aria-disabled` when the Toggle is disabled 

## Test instructions

- go in the standalone app (yarn start go to localhost:3333)
- go in the new "UI Controls" tab
- verify the Toggle and CornerstoneToggle demos work as expected
- switch to RTL pressing the button at the top of the page
- verify again the Toggle and CornerstoneToggle demos work as expected

Also, please verify the style adjustments from #682 are OK.

Fixes #681


andizer:
Acceptance done 👍  I cannot merge this pull because there are some failing tests. I have no clue how to fix these.

PR #700, opened 2018-08-07
Adds HelpText component

### LINKED ISSUE #694: Help text: fix a11y and props
By default, the help text uses a gray color `$color_grey_medium_dark` set to `#888888`. The contrast ratio against a white background is 3.54:1 which is acceptable just for "large text" (24 pixels or 18.5 pixels bold).

For normally sized text, the contrast ratio should be at least 4.5:1 so the lightest gray that can be used is `$color_grey_text` set to `#646464`.

- the prop `textColor` is OK but the CSS property should be `color`, not `textColor`
- the default font-size is `0.8em` which computes to 12.8 pixels in the standalone app but in the plugin it's only 10.4 pixels because the WordPress uses 13 pixels as base font size

Xyfi:
## Summary

* Adds `HelpText` component.
* Replaced `ExplanationText` in `SynonymsInput` with `HelpText`.

* Adds new color to `colors.scss`. The minimum gray color that can be used on white accessibility wise.

## Test instructions

*

Fixes #694


boblinthorst:
CR: ok 👍

composites/Plugin/Shared/components/SynonymsInput.js
```diff
@@ -16,9 +12,9 @@ const SynonymsInput = ( { id, label, value, onChange, explanationText } ) => {
 				{ label }
 			</YoastInputLabel>
 			{ explanationText !== null && (
```
afercia: not sure, why not simply use `{ explanationText && (`

boblinthorst: not sure either, changing it.

boblinthorst:
What started as a quick fix in a review turned into rather large code changes 😅

Significant changes are:

- added helptext to the snippet editor as well in its sibling pr.
- Removed font size of .8 em after discussion with @afercia , who discussed this with @hedgefield as well.
- Kept the names of the props. as `color` and `fontSize` as props would be rendered as HTML attributes in the markup

afercia:
SInce the new color will be used as the lightest gray we can use for text against a white background to have a contrast of 4.5:1 can we rename it and improve the comment:

Current:
`$palette_grey_medium_darker: #767676; // Safe to use with $palette_white.`

Better:
`$palette_grey_text_light: #767676; // Lightest gray for text on a white background.`

Xyfi:
Reviewed @boblinthorst's changed :+1:

app/KeywordExample.js
```diff
@@ -88,12 +88,12 @@ export default class KeywordExample extends Component {
 	render() {
 		return (
 			<Container>
-				<HelpText text={ [
-					"Enter the search term you'd like this post to be found with and see how it would rank. ",
+				<HelpText>
+					{ "Enter the search term you'd like this post to be found with and see how it would rank. " },
 					<HelpTextLink key="1" href="https://yoa.st/">
 						Learn more about the Content Analysis Tool.
 					</HelpTextLink>,
```
afercia: minor: trailing commas should be removed on lines 92 and 95

composites/Plugin/SnippetPreview/components/HelpTextWrapper.js
```diff
@@ -121,7 +121,9 @@ class HelpTextWrapper extends React.Component {
 						className={ this.props.className + "__panel" }
 						panelMaxWidth={ this.props.panelMaxWidth }
 					>
-						<HelpText text={ this.props.helpText } />
+						<HelpText>
+							{ this.props.helpText },
```
afercia: this trailing comma should be removed

afercia:
Pushed minor coding standard and minor fixes. Fixed the test. Also, in the example I've removed `showLabel` on the `SynonymsInput` component, as it doesn't do anything.

PR #701, opened 2018-08-08
Avoid Collapsible headers vertical scrollbar in IE 11.

afercia:
## Summary

* Avoid Collapsible headers to show a vertical scrollbar in IE 11.

In IE 11, when rendered in the sidebar, the collapsibles show a vertical scrollbar:

screenshot 49 2

Slightly increasing the line-height on the title container fixes it. 

## Test instructions

- needs to be tested in IE 11
- this would require to yarn-link the components, built the JS, etc. and follow the whole process to test on IE 11
- since it's a minor one, I'd suggest to just test there are no regressions on mac browsers and do a final test on IE 11 when a zip will be ready

Fixes https://github.com/Yoast/wordpress-seo/issues/10585


boblinthorst:
Acceptance: It doesn't break anything new in other browsers. For testing in ie11, we need to create a beta of this components branch in wpseo's release/8.0 branch. Which is something i was having some issues with, so i set this issue aside for now.

PR #703, opened 2018-08-09
Sidebar styling tweaks

dutchmartin:
## Summary

* Changed the collapsible border color to the one gutenberg uses.
* Changed the titles of the analytics to be not bold but normal and to set the text color to black/ very dark grey .
* Removed the max-width from the analysis component to remove the wrapping of the analysis results on big screens
* Added a dropshadow to the input fields in the sidebar.

* Adding a color in the `colors.json`.

## Test instructions

* Checkout this branch and link it into [wordpress seo](https://github.com/Yoast/wordpress-seo).
* Go to a post and edit it.
* Verify the sidebar looks like the image below and the changelog entries are correct.
cat___ _qwerty_ _wordpress

Fixes https://github.com/Yoast/wordpress-seo/issues/10596


composites/Plugin/Shared/components/Collapsible.js
```diff
@@ -65,6 +66,7 @@ const StyledTitle = styled.span`
 	white-space: nowrap;
 	text-overflow: ellipsis;
 	overflow-x: hidden;
+	color: ${ colors.$palette_grey_dark };
```
IreneStr: Please never use a `$palette_...` color. Use `$color_...` instead. In this case: `$color_headings`

style-guide/_colors.scss
```diff
@@ -2,6 +2,7 @@ $palette_white:              #fff;    // Safe to use with $palette_grey_text –
 // Shades of gray from lighter to darker.
 $palette_grey_ultra_light:   #f7f7f7; // Safe to use with $palette_grey_text – $palette_black.
 $palette_grey_light:         #f1f1f1; // Safe to use with $palette_grey_text – $palette_black.
+$palette_grey_medium_light:  #e2e4e7;
```
Dieterrr: Did you discuss adding this color to the style guide with design? If not, please first make sure they are ok with it, otherwise use one of the existing colors from the style guide.

IreneStr: This has been discussed with @hedgefield and he was fine with it 😄 But good question to ask!

IreneStr:
Acceptance 🚧 

- You've introduced `palette_grey_medium_light` but are not using it. The dividers are still too light grey.
- You've introduced an `!important` in the CSS. Please ask for help to find a different solution. @afercia might be your best bet. Only when nothing else seems to work, use`!important`.
-  The focus keyword input field in the sidebar is styled completely differently than the other input fields: they have rounded corners and the border is too dark. 

For the next tester: don't forget to test the multiple keyword input fields in premium.

composites/Plugin/Shared/components/YoastInput.js
```diff
@@ -18,7 +18,11 @@ export const YoastInputLabel = styled.label`
 `;
 
 export const YoastInputField = styled.input`
-	border: 1px solid ${ colors.$color_input_border };
-	padding: 0.75em;
-	font-size: 1em;
+	&&& {
+		box-shadow: inset 0 1px 2px rgba(0,0,0,.07);
```
afercia: from the style-guide readme:
> do not use rgba colors in the palette, instead use the `rgba()` function

I'd agree the wording of this recommendation could be improved. However, it's meant to not introduce in the codebase colors that are not part of the palette. In this case it's a minor issue because it's black but for consistency I'd recommend to use `${ rgba( colors.$color_black, 0.7 ) }`

composites/Plugin/Shared/components/YoastInput.js
```diff
@@ -18,7 +18,11 @@ export const YoastInputLabel = styled.label`
...
+		border: 1px solid ${ colors.$color_input_border };
+		padding: 0.75em;
```
afercia: I know this value is pre-existing, but it was discussed at some point because `em` values don't make much sense in this context and don't match our 8 pixels based grid size. Both in the standalone app and in Gutenberg the computed value is `9.75px` which will also be subject to browsers rounding. I'd consider to use a different value, in pixels /Cc @hedgefield

composites/Plugin/Shared/components/YoastInput.js
```diff
@@ -18,7 +18,11 @@ export const YoastInputLabel = styled.label`
...
+		font-size: 13px;
```
afercia: I'd keep it to `1em` (which is equivalent to 13px in Gutenberg) unless there are specific reasons to use a value in pixels.

afercia:
CR 🚧 
See a few comments, mostly related to best practices.

afercia:
Right now, this can't be tested on the plugin because:
- the plugin trunk still has the `getI18n` error
- the `hotfix/8.3.1` branch that fixes the `getI18n` error has a wrong `WPSEO_VERSION` that prevents to build assets correctly
to actually test it, either the first or the second need to be manually fixed temporarily

composites/Plugin/Shared/components/YoastInput.js
```diff
@@ -18,7 +19,11 @@ export const YoastInputLabel = styled.label`
 `;
 
 export const YoastInputField = styled.input`
-	border: 1px solid ${ colors.$color_input_border };
 	padding: 0.75em;
 	font-size: 1em;
+	&&& {
+		box-shadow: inset 0 1px 2px ${ rgba( colors.$color_black, 0.07 ) };
+		border: 1px solid ${ colors.$color_input_border };
+		border-radius: 0;
```
afercia: Noticed that setting `box-shadow` and `border` to a so high specificity prevents the focus style to work properly in the metabox. Instead in the sidebar the focus style is inherited by Gutenberg and it works.

afercia:
CR 👍  

Pushed a couple changes to:
- set the input field padding to 8 pixels as per design feedback
- fix the input field focus style

Will split the hardcoded colors in a separate issue.

When doing acceptance, please verify the fields (keyphrase, additional keyphrase, synonyms) have the same styling in the metabox and in the sidebar, including the focus style:

screen shot 2018-10-01 at 11 32 07

IreneStr:
Acceptance:

- [ ] The dividers between the sidebar sections should have the same color as Gutenberg uses, i.e. a darker grey.

- [ ] The titles within the analysis should have the same font color as the analysis descriptions (= black/ very dark grey) and shouldn't be bold.
- [ ] All input fields in the sidebar (keywords, synonyms, additional focus keyword (more?)) should have the same style as the input fields in the snippet editor. This means that at least a dropshadow should be added).
- [ ] Remove the max-width from the analysis component to remove the wrapping of the analysis results on big screens.

PR #741, opened 2018-09-19
Add the tests to the eslint config and include every js file by default.

### LINKED ISSUE #737: Don't exclude tests from the ESLint config
We shouldn't exclude test files from being linted:

https://github.com/Yoast/yoast-components/blob/4d5f52ecfde449a2f0a3fd97d598d8667a1f0f58/Gruntfile.js#L14-L20

dutchmartin:
## Summary

* N/A

* Including every .js file by default in ESlint. 
* Fixed various errors that occurred by deleting code after discussing it with @atimmer .

## Test instructions

* Run the unit tests by executing `yarn test ` and eslint by executing `grunt eslint`

Fixes #737


IreneStr:
CR: ❓ 
Some questions: 
- Why are app/ and redux/ excluded from the linting?
- In `composites/OnboardingWizard/tests/wizardTest.js`,  the`has correct initial state` test has been removed. Is this intentional?
- In the same file, there are some empty tests. If you're already removing non-functioning test, please remove those as well.

dutchmartin:
@IreneStr 
1. Those where generating a lot of errors, so i decided to not fix them since that would broaden up the scope of this issue to fix all code style issues in this repo. The folders are excluded because travis executes eslint and fails on errors, which would prevent this from being merged.
2. Yes, that is exactly what i meant by `Fixed various errors that occurred by deleting code after discussing it with @atimmer.` in the PR description.
3. This feels like [scope creep](https://en.wikipedia.org/wiki/Scope_creep) to me, but I will do it after  redefinening the scope of this issue.

dutchmartin:
Please note that some things are fixed by eslint autofix. If there are any weird things that happend with that, the eslint config may have to be modified.

PR #743, opened 2018-09-24
Fix import in the Snippet Preview demo.

### LINKED ISSUE #740: Components standalone is broken
The components standalone is broken:
schermafbeelding 2018-09-19 om 10 04 50

To reproduce: checkout develop, run `yarn install`, run `yarn start` and go to `localhost:3333`

afercia:
Fixes the snippet preview demo in the standalone app.

To test:
- check the standalone app: `yarn start` and go to `localhost:3333`
- go in the Snippet preview tab
- verify there are no errors and the snippet preview / editor work as expected
- edit the meta description and verify the progress bar works as expected (max length is 156 characters)

Note: since the date in the meta description demo is hardcoded (`Jan 8, 2018 -`) you can actually enter only 143 characters before the progress bar turns into orange.

Fixes #740

### DIFF
--- a/app/SnippetEditorExample.js
+++ b/app/SnippetEditorExample.js
@@ -3,7 +3,8 @@ import React, { Component } from "react";
 import styled from "styled-components";
 import debounce from "lodash/debounce";
 
-import MetaDescriptionLengthAssessment from "yoastseo/src/assessments/seo/metaDescriptionLengthAssessment";
+import { assessments } from "yoastseo";
+const { MetaDescriptionLengthAssessment } = assessments.seo;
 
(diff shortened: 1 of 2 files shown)


IreneStr:
CR 🚧 We don't want to import yoastseo.js stuff from the node_modules anymore. It should use the global instead.

afercia:
So as far as I see this requires a PR on YoastSEO.js as `MetaDescriptionLengthAssessment` is not in the global exports.

IreneStr:
CR 👍

IreneStr:
Acceptance 👍

PR #749, opened 2018-10-03
Added id to draft js editor field

### LINKED ISSUE #637: Implement custom fillFieldDraftJS and fillClearFieldDraftJS for automated testing
As agreed with @atimmer the following functions _fillFieldDraftJS_ and _fillClearFieldDraftJS_ are to be introduced in [automated testing](https://github.com/Yoast/plugins-automated-testing). The issue is created here in order to be picked up in the process. The idea is to help me to write this function so I can test it immediately in the automated test scenarios.

More details: fillField is currently implemented in codeceptjs as follows:
https://github.com/Codeception/CodeceptJS/blob/master/lib/helper/Puppeteer.js#L872
```  async fillField(field, value) {
    const els = await findFields.call(this, field);
    assertElementExists(els, field, 'Field');
    const el = els[0];
    // await el.focus();
    const tag = await el.getProperty('tagName').then(el => el.jsonValue());
    const editable = await el.getProperty('contenteditable').then(el => el.jsonValue());
    if (tag === 'INPUT' || tag === 'TEXTAREA') {
      await this._evaluateHandeInContext(el => el.value = '', el);
    } else if (editable) {
      await this._evaluateHandeInContext(el => el.innerHTML = '', el);
    }
    await el.type(value, { delay: 10 });
    return this._waitForAction();
  }
```

It is proposed to implement custom fillFieldDraftJS based on fillField locating element for example within "DraftEditor-root".
The same is supposed to be implemented for fillClearFieldDraftJS based on clearField of codeceptjs.

Xyfi:
## Summary

* Adds an id to the draft JS editor `editablecontent` element.

*

## Test instructions

* Make sure this branch is checked out and `yoast-components` is linked to `wordpress-seo`.
* On the post edit page open the snippet editor.
* Inspect the source and look for the div with id `snippet-editor-field-title`. This was the existing id on the container wrapper.
* Now look for the div with id `snippet-editor-field-title-field`. This is the newly added id. Make sure the same div has the `contenteditable` attribute.
* The same can be done for `snippet-editor-field-description` and `snippet-editor-field-title-field`.

Fixes #637


composites/Plugin/SnippetEditor/components/ReplacementVariableEditorStandalone.js
```diff
@@ -345,9 +346,22 @@ class ReplacementVariableEditorStandalone extends React.Component {
 	 * @returns {void}
 	 */
 	setEditorRef( editor ) {
+		try {
+			editor.editor.editorContainer.firstChild.id = this.props.fieldId;
```
jcomack: `editor.editor.editorContainer` seems like a strange naming issue. Perhaps renaming the passed argument would be a good idea.

composites/Plugin/SnippetEditor/components/ReplacementVariableEditor.js
```diff
@@ -99,6 +99,7 @@ class ReplacementVariableEditor extends React.Component {
 					id={ fieldId }
 				>
 					<ReplacementVariableEditorStandalone
+						fieldId={ `${ fieldId }-field` }
```
jcomack: Suffixing a field with `-field` seems a little strange in terms of naming, especially when you look at the snapshots, which contain fields that have the word 'field' in there twice. Example: `snippet-editor-field-title-field`.

composites/Plugin/SnippetEditor/components/ReplacementVariableEditorStandalone.js
```diff
@@ -345,9 +346,22 @@ class ReplacementVariableEditorStandalone extends React.Component {
...
+		} catch ( error ) {
+			console.trace( error );
+		}
+
 		this.editor = editor;
 	}
 
+	setEditorFieldId() {
+		const editorContainer = get( this.editor, "editor.editorContainer" );
+		const editorField = editorContainer.querySelector( "div[contenteditable]" );
```
jcomack: Unsure about this, but I believe the equal signs need to be aligned if we want to adhere to the codestyle.

composites/Plugin/SnippetEditor/components/ReplacementVariableEditor.js
Xyfi: We can't rename the earlier id we introduced because of backwards compatibility.

Dieterrr:
CR done 👍

afercia:
I must assume the test instructions haven't been updated and this concern:
> We can't rename the earlier id we introduced because of backwards compatibility. The snippet-editor-field-title part comes from a prop, so it can be changed if the developer that uses the component chooses too.

has since then been reconsidered. If so, please let me know and this can be merged 🙂 /Cc @Xyfi

afercia:
Discussed internally and seems the ID renaming is not a concern.

PR #753, opened 2018-10-10
Fix snippet preview slug

jcomack:
## Summary

* Fixes a bug where categories weren't properly reflected in the Snippet Preview's slug.

*

## Test instructions

* See https://github.com/Yoast/wordpress-seo/pull/11210 for test instructions.

### DIFF
--- a/composites/Plugin/SnippetEditor/components/SnippetEditor.js
+++ b/composites/Plugin/SnippetEditor/components/SnippetEditor.js
@@ -451,9 +451,15 @@ class SnippetEditor extends React.Component {
 	 * @returns {string} The formatted permalink.
 	 */
 	buildPermalink( url, originalData ) {
-		let slug      = Object.values( originalData.parents ).join( "/" ) + "/" + originalData.primaryTaxonomySlug;
-		let permalink = url.replace( "%category%", slug );
-		permalink     = permalink.replace( "%postname%", originalData.slug );
+		let { slug, parents = [], primaryTaxonomySlug = "" } = originalData;
+
+		if ( originalData.parents ) {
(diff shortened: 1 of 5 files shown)


IreneStr:
CR by @andizer  done 👍

composites/Plugin/SnippetEditor/components/SnippetEditor.js
```diff
@@ -414,34 +415,58 @@ class SnippetEditor extends React.Component {
...
+	buildPermalink( url, originalData ) {
+		let parents = [];
+		const { slug, primaryTaxonomySlug = "" } = originalData;
+
+		if ( originalData.parents ) {
+			parents = Object.values( originalData.parents );
+		}
+
+		const categorySlug = [ ...parents, primaryTaxonomySlug ].join( "/" );
+		let permalink      = url.replace( "%category%", categorySlug );
```
andizer: This is WordPress specific. Is there a way to move this to the Yoast SEO plugin?

atimmer:
Talked to @jcomack about this. I think `yoast-components` should have some logic for URLs. But only the CMS-agnostic parts. So I propose passing the following to `yoast-components` from the plugin:

```js
{
  structure: "siteurl.nl/[category]/[postname],
  data: {
    category: "my-primary-category",
    postname: "my-postname",
  }
}
```

Then if `category` or `postname` change, the new data can be passed. This will also work for other CMSs.

PS. In my example I write `[]` around placeholder, for ease of implementation this can also be `%%`. This doesn't really change the API that much. `[]` is just clearer as an CMS-agnostic example.

PR #758, opened 2018-10-15
Update SynonymsInput to use a help link.

afercia:
## Summary

* N/A

Updates the SynonymsInput to not use an help text any longer and use a help link icon instead.

Fixes https://github.com/Yoast/wordpress-seo-premium/issues/2058

### DIFF
--- a/composites/Plugin/Shared/components/SynonymsInput.js
+++ b/composites/Plugin/Shared/components/SynonymsInput.js
@@ -1,21 +1,30 @@
 import React from "react";
 import PropTypes from "prop-types";
 import uniqueId from "lodash/uniqueId";
+import styled from "styled-components";
 
 import { YoastInputContainer, YoastInputField, YoastInputLabel } from "./YoastInput";
-import HelpText from "../../Shared/components/HelpText";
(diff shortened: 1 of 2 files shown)


composites/Plugin/Shared/components/SynonymsInput.js
```diff
@@ -1,21 +1,30 @@
...
-			{ explanationText && (
-				<HelpText>
-					{ explanationText }
-				</HelpText>
-			) }
+			<SynonymsFieldLabelContainer>
+				<StyledYoastInputLabel htmlFor={ id }>
+					{ label }
+				</StyledYoastInputLabel>
+				{ labelSiblingElement }
```
jcomack: Is this element ever going to be something else than a help link? If not, the name `labelSiblingElement` is too abstract.

afercia: Thought to use this name for consistency with `KeywordInput` where the prop was previously named `postLabelElement`. No objections to change it, then it should be changed in `KeywordInput` too. Suggestions on the name welcome 🙂

Dieterrr:
Acceptance done 👍

PR #798, opened 2018-12-24
Add Underline and outboundlink to header. 

Dieterrr:
## Summary

* N/A

* Made link on Courses card header outbound and added underline on header title. 

## Test instructions

* Yarn link yoast-components in this branch and build yoastseo. 
* Go to the courses page click on the header (image or title) and see the link opens in a new tab.
* Hover over the header and see the title gets an underline. 

Fixes #[11906](https://github.com/Yoast/wordpress-seo/issues/11906)

### DIFF
--- a/composites/CoursesOverview/Card.js
+++ b/composites/CoursesOverview/Card.js
@@ -37,6 +37,7 @@ const HeaderLink = styled.a`
 		color: ${ colors.$color_pink_dark };
 	}
 
+	&:focus,
 	&:active {
 		box-shadow: none;
 	}
--- a/composites/CoursesOverview/Card.js
+++ b/composites/CoursesOverview/Card.js
@@ -61,7 +61,7 @@ class Card extends React.Component {
 
 		if ( this.props.header.link ) {
 			return (
-				<OutboundHeaderLink href={ this.props.header.link }>
+				<OutboundHeaderLink href={ this.props.header.link } rel={ null }>
 					<HeaderImage src={ this.props.header.image } alt="" />
 					<HeaderTitle>{ this.props.header.title }</HeaderTitle>
 				</OutboundHeaderLink>
--- a/composites/CoursesOverview/Card.js
+++ b/composites/CoursesOverview/Card.js
@@ -38,9 +38,9 @@ const HeaderTitle = styled.h2`
 
 const HeaderLink = styled.a`
 	text-decoration: none;
-	
+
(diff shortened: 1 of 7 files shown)


composites/CoursesOverview/Card.js
```diff
@@ -55,10 +64,10 @@ class Card extends React.Component {
 
 		if ( this.props.header.link ) {
 			return (
-				<HeaderLink href={ this.props.header.link }>
+				<OutboundHeaderLink href={ this.props.header.link }>
```
afercia: we should use `rel={ null }` to omit noopener and noreferrer, as these links point to yoast.com

afercia:
CR 👍 

Pushed a few changes:
- use empty alt attributes for the images, as recommended in https://www.w3.org/TR/WCAG20-TECHS/H2.html the link text is already announced, no need to repeat it in the imate alt
- added the courses overview in the standalone examples, under the "Components" tab
- RTL styles for the checkmark icon
- CSS clean up and standardize the buttons effect on :active
- use `rel={ null }` to omit the `rel="noopener noreferrer"` attributes, as these are links that point to yoast.com

To test the RTL styles:
- go in the standalone app: yarn start and go to http://localhost:3333/
- go in the "Components" tab and press the "Change language direction" button 
- verify the checkmarks position is inverted and the Cards look OK

IreneStr:
Acceptance 🚧 
- The underlinings in safari are blue instead of pink/purple:
schermafbeelding 2019-01-02 om 12 05 42
- UX question @afercia @hedgefield: the title is also underlined when the image is hovered. Is that intentional/desirable?

hedgefield:
I suppose that's okay as they both lead to the same place. But whether it's intentional, I don't know.

afercia:
Last commit:
- fixes hover style in Safari
- resets the blue box-shadow _while_ clicking

IreneStr:
Acceptance 👍 I've pushed one minor change to prevent the blue box-shadow after clicking in Chrome.

afercia:
> I've pushed one minor change to prevent the blue box-shadow after clicking in Chrome.

Worth noting this change also removes the focus style when using the keyboard: the only indicator now is the underline on the text, which is not great but at least it's something. The persistence of the style "after clicking" is expected, as at that point the link is no more active but it does have focus.

We should try to avoid to change the focus style when the change is not an accessibility improvement 🙂

PR #800, opened 2019-01-02
Make getFeed work in IE11 and Edge

afercia:
## Summary

* Fixes parsing of feeds in Internet Explorer 11

- uses [Wicked Good XPath](https://github.com/google/wicked-good-xpath) for browsers that don't support XPath evaluation (IE11), for more details see their documentation. I'm not an expert in XPath parsing in JS, but it works. /Cc @atimmer @herregroen 
- adds a real dashboard widget working example in the demos

## Test instructions

Updated instructions:

- at the moment, this PR can't be tested in the standalone app because of https://github.com/Yoast/YoastSEO.js/issues/2086
- if you want to test it there, you need to downgrade `yoastseo` to `v1.43.0` in the package.json

## Test in the plugin
- first, test on `release-yoast-seo/9.5`
- go to the WP dashboard and observe in IE11 the Yoast dashboard widget doesn't show the posts from the feed
- go in the Courses page and observe in IE11 the page is blank
- in Edge, the Courses page shows the cards with missing content
- see in the IE11 console the error `[object Error]` and expand it for more details: `TypeError: Object doesn't support property or method 'createNSResolver'`

Switch the plugin to the `11935-get-feed-ie11-edge` branch (see https://github.com/Yoast/wordpress-seo/pull/11968) and yarn link this yoast-component PR to the plugin:
- build the JS: `grunt build:js && grunt webpack:buildProd`
- verify the dashboard widget shows the content parsed from the feed
- verify the "Read more like this ..." link at the bottom has a link to yoast.com with tracking code
- you may need to flush the transients, if you're using VVV:
```
vagrant ssh
cd /srv/www/wordpress-develop/public_html/
wp transient delete-all
```

- verify the Courses page shows the content parsed from the feed in IE11 and Edge
- in both the dashboard widget and the Courses page, verify all the links that have `target="_blank"` don't have a `rel="noopener noreferrer" attribute` and do have the a11y message "Opens in a new. browser tab"

Note: the Courses page layout is broken in IE11 because the current CSS grid implementation doesn't work in this browser. Will be addressed separately.

Screenshots from IE11:

screenshot 160

screenshot 161

(as said, the broken layout will be addressed in https://github.com/Yoast/wordpress-seo/issues/11953 )

Fixes https://github.com/Yoast/wordpress-seo/issues/11935
Fixes https://github.com/Yoast/wordpress-seo/issues/11073

### DIFF
--- a/app/DashboardWidgetWrapper.js
+++ b/app/DashboardWidgetWrapper.js
@@ -1,6 +1,8 @@
 import React from "react";
 import styled from "styled-components";
 
+import getPostFeed from "../utils/getPostFeed";
+
 import {
 	DashboardWidgetSeoAssessment,
 	DashboardWidgetWordpressFeed,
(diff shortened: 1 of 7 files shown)


andizer:
CR Done 👍

afercia:
Seems there are additional issues with Edge, where each feed item get parsed but many of the item elements are undefined (which can also explain why the cars look "empty" in Edge, see https://github.com/Yoast/wordpress-seo/issues/11953#issuecomment-450953569):

screenshot 171

At the moment, I have no idea what's going on.

afercia:
Moving back to development as the Edge issue needs more investigation, and some help 🙂

afercia:
Re: the Edge problem: after an insane amount of time spent investigating on this, turns out in Edge `getElementsByTagName` doesn't recognize namespaces, at least _in some cases_. It was introduced here to check for the existence of an element: https://github.com/Yoast/yoast-components/commit/aa87a129c6f0bfc2aff9c4a542f1693636bc11e6

What's happening is:
- in the Courses feed, Edge doesn't recognize all the elements with a colon (namespace) in their name, for example: `<content:encoded>`
- removing the `getElementsByTagName` check makes things work again
- what is very weird though, is that in the dashboard widget feed, Edge is perfectly able to get `<content:encoded>`
- I've verified this by just replacing the feed URL in the courses overview
- my guess is there's a difference in the two feeds that triggers _something_ in Edge, suspect: the declared xml namespaces at the top of the feed are different

afercia:
Additionally, in the WordPress dashboard widget the last link has no `href` as seems `getFeedMeta()` is not able to parse `/rss/channel/link` (not even `/rss/channel/title` and `/rss/channel/description`):

screenshot 2019-01-04 at 10 27 36

app/DashboardWidgetWrapper.js
```diff
@@ -18,58 +20,88 @@ export const DashboardContainer = styled.div`
...
+				{
+					value: 47,
+					color: "#0F0",
+					html: "Posts with a <b>good</b> score",
+				},
+			],
+			feed: null,
+		};
+
+		this.getFeed( "free" );
```
herregroen: Why is an argument passed to a function that takes no arguments?

app/DashboardWidgetWrapper.js
```diff
@@ -18,58 +20,88 @@ export const DashboardContainer = styled.div`
...
+					item.description = item.description.replace( `The post ${ item.title } appeared first on Yoast.`, "" ).trim();
+					item.content = item.content;
+
+					return item;
+				} );
+
+				this.setState( { feed } );
+			} )
+			/* eslint-disable-next-line no-console */
+			.catch( error => console.log( error ) );
```
herregroen: Probably better to make this a `console.error`.

utils/getFeed.js
```diff
@@ -107,6 +109,10 @@ function getFeedItems( parsed, nsResolver, maxItems, parseFeedItem ) {
 export function parseFeed( raw, maxItems = 0, parseFeedItem ) {
 	return new Promise( function( resolve, reject ) {
 		try {
+			// Use Wicked Good XPath for browsers that don't support XPath evaluation.
+			if ( "evaluate" in document === false ) {
```
herregroen: For this PR I think this is the way to go. Especially since we're also including `fetch` as a dependency.

But we should definitely consider expecting these things to simply exist. `wp_get_script_polyfill` could easily be expanded to conditionally load `wicked-good-xpath` only when necessary.

afercia:
Last commit re-implements the check for element existence but this time using XPath `count()`.

However, worth noting a few things:

This works if the element doesn't exist or the XPath expression is malformed in the part that represents _the element_ for example:
`getXPathText( "child::content:XYZencoded", parsed, snapshot, nsResolver );`

It doesn't work if the malformed part is the _namespace_ part: it will throw a DOMException, for example:
`getXPathText( "child::XYZcontent:encoded", parsed, snapshot, nsResolver );`

Regardless, I'm not sure why we want to check for the elements existence in the first place. We're passing these elements to React (unless we're using this in other places too): if the element doesn't exist this method was already returning `undefined` (note: I've changed it to `null`) so React won't render anything.

Worth also noting some of the components that use this, have `required` properties: if the element is missing, React will throw a warning. /Cc @herregroen

app/DashboardWidgetWrapper.js
afercia: I'm not sure we should throw a console.error in the users browser. More importantly, the general usage of `console` in the yoast-components should be re-thought, as it's one of the reasons why the standalone app doesn't work in Internet Explorer 11. In fact, in IE11, the console object is not available by default. It is only when the "dev tools" are open. So, with the dev tools closed, all you can see is a blank page.

If it was up to me, I'd opt for removing any usage of `console` from our codebase. Also, there's a reason why ESLint has a `no-console` rule and I'm not sure disabling the rule is a good path forward.

utils/getFeed.js
```diff
@@ -20,10 +22,9 @@ import { parseFeedItem as parsePostFeedItem } from "./getPostFeed";
  * @returns {string|undefined} The string result of the xpath query.
  */
 export function getXPathText( xpath, document, context = null, nsResolver = null ) {
-	const tagname = xpath.replace( "child::", "" );
-	const elementExists = document.getElementsByTagName( tagname );
+	const elementCount = document.evaluate( "count(" + xpath + ")", ( context || document ), nsResolver, XPathResult.ANY_TYPE, null ).numberValue;
```
IreneStr: Please add a comment about what you're doing here/what you're doing here. It's hard to understand what's happening here

app/DashboardWidgetWrapper.js
IreneStr: Whether it's a good idea to use the console at all is another discussion. For now, let's change it to console.error, because it stands out more and might signal a bigger problem in the feed, and therefore in the plugin.

afercia:
Changed `console.log` to `console.error` for both the Dashboard widget and Courses `getFeed` examples. 

Worth noting these examples just re-use the code that's used in wordpress-seo, where `console.log` is still used. Discussed a bit with @IreneStr and maybe it would be better to display some useful message to users when the feeds can't be parsed. This should be addressed separately though.

IreneStr:
2nd CR 👍

Dieterrr:
Acceptance done 👍 
Everything works as advertised :) Styling indeed still broken in IE, in Edge is fine. 

@afercia Thanks for the very clear PR and instructions!

PR #818, opened 2019-02-01
Introduces a FacebookImage component

### LINKED ISSUE #812: Create a FacebookImage component
The FacebookImage component will be rendered as an `img`
yoast_-_home
_Example of the preview with landscape image._

## Properties
* src (full path to the image)
* type (square, portrait or landscape) 

The type is relevant for the way we display the image. See the specifications below.

## Specifications
An image has a minimum width of 500px to be a landscape image. When the ratio is 1:1 it will be a square image. 
When portrait the max-height will be 236 pixels. If the width is lower than 158 px a square image is rendered.

## Styling
The image will be placed in a container:
```css
  line-height: 0;
  position: relative;
  z-index: 1;
  overflow: hidden;
  background-position: 50% 50%;
  background-repeat: no-repeat;
  display: block;
height: '{the height of the given image}px'
  width: 500px;
```

In square mode:
```css
    background-color: #fff;
    height: 158px;
    width: 158px;
```

In portrait mode:
```css
    height: 236px;
    width: 158px;
```

### Styling of the image
The image itself will have the following CSS attributes:
```css
     height: 100%;
    min-height: initial;
    width: 100%;
    position: relative;
```

In square mode:
```css
    height: auto;
```

Parent #814 814

IreneStr:
## Summary

* Introduces a `FacebookImage` component.

* Because what image mode (portait, landscape, square) Facebook is using for which image dimensions is very unclear, we went for a pragmatic solution that closely resembles our current/old implementation:
    - If the height of an image is greater than its width, we use portrait.
    - If the height and width are the same, we use square.
    - If the width is greater than its with, we use landscape.
    - If the height and/or width of the image is less than 158 px, we show an "The image you selected is too small for Facebook" error.
* There are two types of errors: one when the height and/or width of the uploaded image is less than 158 px, and one when the image is no image/can't be used for another reason.
* In contrast to the current social previews, I'm using a shade of red from colors.json.
* I copied the error messages from the current social previews.

## Test instructions

* Run yarn start to start the example. 
* Go to the Facebook tabs.
* You can change the URL in App.js to test some more with other images.

## UI changes
* [ ] This PR changes the UI in the plugin. I have added the 'UI change' label to this PR.

Fixes #812


composites/Plugin/SocialPreviews/Facebook/components/FacebookImage.js
```diff
@@ -0,0 +1,144 @@
...
+const PlaceholderImage = styled.div`
+	height: 261px;
+	width: 500px;
+	background-color: ${colors.$color_grey};
+`;
+
+/**
+ * Renders the FacebookImage component.
+ *
+ * @param {string} src The image source
```
andizer: Full stop missing.

composites/Plugin/SocialPreviews/Facebook/components/FacebookImage.js
```diff
@@ -0,0 +1,144 @@
...
+	/**
+	 * Renders the FacebookImage.
+	 *
+	 * @returns {ReactComponent} Either the ErrorImage component or the FacebookImageContainer.
+	 */
+	render() {
+		const imageProperties = this.state.imageProperties;
+		const status = this.state.status;
+
+		if ( status === "loading" ) {
```
andizer: You might consider having a constant containing the value `loading`

composites/Plugin/SocialPreviews/Facebook/components/FacebookImage.js
```diff
@@ -0,0 +1,144 @@
...
+			return <PlaceholderImage />;
+		}
+
+		if ( status === "errored" ) {
```
andizer: See my comment above. Have a constant containing the value `errored`

composites/Plugin/SocialPreviews/Facebook/components/FacebookImage.js
```diff
@@ -0,0 +1,144 @@
...
+	/**
+	 * Gets the dimensions for the Facebook image container.
+	 *
+	 * @param {string} imageMode The facebook image mode: either landscape, square or portrait.
+	 *
+	 * @returns {Object} The width and height for the container.
+	 */
+	getContainerDimensions( imageMode ) {
+		switch ( imageMode ) {
+			case "landscape":
```
andizer: I would love to see a constant for the value `landscape`

composites/Plugin/SocialPreviews/Facebook/components/FacebookImage.js
```diff
@@ -0,0 +1,144 @@
...
+				return {
+					height: "261px",
+					width: "500px",
+				};
+			case "square":
```
andizer: Maybe a constant for `square`

composites/Plugin/SocialPreviews/Facebook/components/FacebookImage.js
```diff
@@ -0,0 +1,144 @@
...
+				return {
+					height: "158px",
+					width: "158px",
+				};
+			case "portrait":
```
andizer: Maybe a constant for `portrait`

composites/Plugin/SocialPreviews/Facebook/helpers/determineFacebookImageProperties.js
```diff
@@ -0,0 +1,172 @@
...
+/**
+ * Determines the image display mode of the Facebook Image.
+ *
+ * @param {Object} dimensions The image's dimensions.
+ *
+ * @returns {string} The display mode of the image.
+ */
+function determineFacebookImageMode( dimensions ) {
+	if ( dimensions.height > dimensions.width ) {
+		return "portrait";
```
andizer: I think make a constant of this would be nice :)

composites/Plugin/SocialPreviews/Facebook/helpers/determineFacebookImageProperties.js
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	if ( dimensions.height === dimensions.width ) {
+		return "square";
```
andizer: See my earlier constant comments

composites/Plugin/SocialPreviews/Facebook/helpers/determineFacebookImageProperties.js
```diff
@@ -0,0 +1,172 @@
...
+	 */
+	if ( imageMode === "square" ) {
+		return {
+			width: squareWidth,
+			height: squareHeight,
+		};
+	}
+
+	/*
+	 When we're in portrait or landscape mode, we can't just resize to the dimensions expected by Facebook. If we'd do
```
IreneStr: I'll move these docs

composites/Plugin/SocialPreviews/Facebook/helpers/determineFacebookImageProperties.js
```diff
@@ -0,0 +1,172 @@
...
+function determineFacebookImageMode( dimensions ) {
+	if ( dimensions.height > dimensions.width ) {
+		return "portrait";
+	}
+
+	if ( dimensions.height === dimensions.width ) {
+		return "square";
+	}
+
+	return "landscape";
```
andizer: See my earlier constant comments

composites/Plugin/SocialPreviews/Facebook/components/FacebookImage.js
```diff
@@ -0,0 +1,144 @@
...
+	getContainerDimensions( imageMode ) {
+		switch ( imageMode ) {
+			case "landscape":
+				return {
+					height: "261px",
+					width: "500px",
+				};
+			case "square":
+				return {
+					height: "158px",
```
andizer: You can export the constants in https://github.com/Yoast/yoast-components/pull/818/files#diff-caec132fef457c99862b77535af59077R1 and use them here.

composites/Plugin/SocialPreviews/Facebook/components/FacebookImage.js
```diff
@@ -0,0 +1,144 @@
...
+		const status = this.state.status;
+
+		if ( status === "loading" ) {
+			return <PlaceholderImage />;
+		}
+
+		if ( status === "errored" ) {
+			return <ErrorImage>The given image url cannot be loaded</ErrorImage>;
+		}
+		if ( imageProperties.height < 158 || imageProperties.width < 158 ) {
```
andizer: You might consider using the constants declare here: https://github.com/Yoast/yoast-components/pull/818/files#diff-caec132fef457c99862b77535af59077R1

composites/Plugin/SocialPreviews/Facebook/components/FacebookImage.js
```diff
@@ -0,0 +1,144 @@
...
+`;
+
+const FacebookImageContainer = styled.div`
+	line-height: 0;
+	position: relative;
+	z-index: 1;
+	overflow: hidden;
+	display: block;
+	background-color: #fff;
+	height: ${ props => props.dimensions.height };
```
andizer: Maybe you can add `px` here

composites/Plugin/SocialPreviews/Facebook/components/FacebookImage.js
```diff
@@ -0,0 +1,144 @@
...
+		}
+	}
+
+	/**
+	 * Renders the FacebookImage.
+	 *
+	 * @returns {ReactComponent} Either the ErrorImage component or the FacebookImageContainer.
+	 */
+	render() {
+		const imageProperties = this.state.imageProperties;
```
andizer: (Maybe) you can do `const { imageProperties, status } = this.state`

composites/Plugin/SocialPreviews/Facebook/helpers/determineFacebookImageProperties.js
```diff
@@ -0,0 +1,172 @@
...
+ * Gets the ratios of the width and height of the original image in relation to the width and height
+ * of the expected image.
+ *
+ * @param {Object} dimensions The dimensions of the original image.
+ * @param {string} imageMode The image mode: either portrait or landscape.
+ *
+ * @returns {Object} The image's width ratio and height ratio.
+ */
+function getImageRatios( dimensions, imageMode ) {
+	if ( imageMode === "portrait" ) {
```
andizer: Constant...

composites/Plugin/SocialPreviews/Facebook/helpers/determineFacebookImageProperties.js
```diff
@@ -0,0 +1,172 @@
...
+		return {
+			widthRatio: dimensions.width / portraitWidth,
+			heightRatio: dimensions.height / portraitHeight,
+		};
+	}
+	if ( imageMode === "landscape" ) {
```
andizer: Constant..

composites/Plugin/SocialPreviews/Facebook/helpers/determineFacebookImageProperties.js
```diff
@@ -0,0 +1,172 @@
...
+ * Calculates the dimensions of the image to use as FacebookImage.
+ *
+ * @param {Object} originalDimensions The width and height of the original image.
+ * @param {string} imageMode The image mode: square, portrait or landscape.
+ *
+ * @returns {object} The image dimensions for the facebook image.
+ */
+function calculateFacebookImageDimensions( originalDimensions, imageMode ) {
+	// Images that are too small should not be scaled.
+	if ( originalDimensions.width < 158 || originalDimensions.height < 158 ) {
```
andizer: You can use `squareWidth` and `squareHeight` here

composites/Plugin/SocialPreviews/Facebook/helpers/determineFacebookImageProperties.js
```diff
@@ -0,0 +1,172 @@
...
+			width: originalDimensions.width,
+			height: originalDimensions.height,
+		};
+	}
+
+	/*
+	 If the image is a square, just use the squareWidth and squareHeight. We don't have to fear
+	 that the resulting image will be warped.
+	 */
+	if ( imageMode === "square" ) {
```
andizer: square can be a constant

andizer (changes requested):
CR done.

As you can see I love using constants :)

composites/Plugin/SocialPreviews/Facebook/components/FacebookImage.js
IreneStr: The 'px' part is already in the props.

IreneStr:
I've discussed the constants with @andizer and decided to use constants for the dimensions (specific heights and widths), but not for the statuses (loading, loaded, errored) and modes (square, portrait, landscape).

composites/Plugin/SocialPreviews/Facebook/components/FacebookImage.js
IreneStr: I think 'my' syntax is more readable.

composites/Plugin/SocialPreviews/Facebook/components/FacebookImage.js
```diff
@@ -0,0 +1,154 @@
...
+	SQUARE_HEIGHT,
+	SQUARE_WIDTH,
+} from "../helpers/determineFacebookImageProperties";
+import colors from "../../../../../style-guide/colors.json";
+
+const MIN_IMAGE_WIDTH = 158;
+const MIN_IMAGE_HEIGHT = 158;
+
+const StyledImage = styled.img`
+	min-height: initial;
```
afercia: not sure why min-height is explicitly set to initial

composites/Plugin/SocialPreviews/Facebook/components/FacebookImage.js
```diff
@@ -0,0 +1,154 @@
...
+	height: ${ props => props.mode === "landscape" ? "auto" : "100%" };
+	width: ${ props => props.mode === "portrait" ? "auto" : "100%" };
+	position: relative;
```
afercia: Not sure why `position: relative;` is used here

composites/Plugin/SocialPreviews/Facebook/components/FacebookImage.js
```diff
@@ -0,0 +1,154 @@
...
+`;
+
+const FacebookImageContainer = styled.div`
+	line-height: 0;
```
afercia: Not sure this line-height is necessary

composites/Plugin/SocialPreviews/Facebook/components/FacebookImage.js
```diff
@@ -0,0 +1,154 @@
...
+	position: relative;
+	z-index: 1;
```
afercia: Not sure position: relative; and z-index are necessary

composites/Plugin/SocialPreviews/Facebook/components/FacebookImage.js
```diff
@@ -0,0 +1,154 @@
...
+	overflow: hidden;
+	display: block;
```
afercia: This should be removed, as a few lines below there's a `display: flex;`

composites/Plugin/SocialPreviews/Facebook/components/FacebookImage.js
```diff
@@ -0,0 +1,154 @@
...
+	display: flex;
+	justify-content: center;
+`;
+
+const ErrorImage = styled.div`
+	height: 261px;
+	width: 500px;
+	text-align: center;
+	box-sizing: border-box;
+	padding-top: 7em;
```
afercia: this could be improved to better vertically align the text, also taking into consideration longer translations

composites/Plugin/SocialPreviews/Facebook/components/FacebookImage.js
```diff
@@ -0,0 +1,154 @@
...
+	font-size: 1rem;
+	color: ${colors.$color_white};
+	background-color: ${colors.$color_red};
```
afercia: missing spaces within brackets in both the colors above

composites/Plugin/SocialPreviews/Facebook/components/FacebookImage.js
```diff
@@ -0,0 +1,154 @@
...
+`;
+
+const PlaceholderImage = styled.div`
+	height: 261px;
+	width: 500px;
+	background-color: ${colors.$color_grey};
```
afercia: missing spaces within brackets

composites/Plugin/SocialPreviews/Facebook/components/FacebookImage.js
```diff
@@ -0,0 +1,154 @@
...
+	render() {
+		const imageProperties = this.state.imageProperties;
+		const status = this.state.status;
+
+		if ( status === "loading" ) {
+			return <PlaceholderImage />;
+		}
+
+		if ( status === "errored" ) {
+			return <ErrorImage>The given image url cannot be loaded</ErrorImage>;
```
afercia: this string is not translatable

composites/Plugin/SocialPreviews/Facebook/components/FacebookImage.js
```diff
@@ -0,0 +1,154 @@
...
+		}
+		if ( imageProperties.height < MIN_IMAGE_HEIGHT || imageProperties.width < MIN_IMAGE_WIDTH ) {
+			return <ErrorImage>The image you selected is too small for Facebook</ErrorImage>;
```
afercia: this string is not translatable

composites/Plugin/SocialPreviews/Facebook/components/FacebookImage.js
```diff
@@ -0,0 +1,154 @@
...
+		}
+
+		const containerDimensions = this.getContainerDimensions( imageProperties.mode );
+		return <FacebookImageContainer mode={ imageProperties.mode } dimensions={ containerDimensions }>
+			<StyledImage src={ this.props.src } mode={ imageProperties.mode } />
```
afercia: - The property `mode` gets rendered in the HTML as an attribute because React thinks it's a valid HTML attribute: it should be renamed.
- the image misses an `alt` attribute: more on this later

composites/Plugin/SocialPreviews/Facebook/components/FacebookImage.js
```diff
@@ -0,0 +1,154 @@
...
+		</FacebookImageContainer>;
+	}
+}
+
+FacebookImage.propTypes = {
+	src: PropTypes.string.isRequired,
```
afercia: if `src` is required, I'm not sure I understand how the initial state works, when users haven't set an image yet

composites/Plugin/SocialPreviews/Facebook/components/FacebookPreview.js
```diff
@@ -15,12 +15,16 @@ import FacebookSiteName from "./FacebookSiteName";
  */
 const FacebookPreview = ( props ) => {
```
afercia: In the docblock: not sure `@constructor` is necessary or correct

composites/Plugin/SocialPreviews/Facebook/components/FacebookPreview.js
```diff
@@ -15,12 +15,16 @@ import FacebookSiteName from "./FacebookSiteName";
...
 	return (
-		<FacebookSiteName siteName={ props.siteName } />
+		<Fragment>
+			<FacebookImage src={ props.src } />
```
afercia: missing `alt` attribute  (see above)

composites/Plugin/SocialPreviews/Facebook/components/FacebookPreview.js
```diff
@@ -15,12 +15,16 @@ import FacebookSiteName from "./FacebookSiteName";
...
+			<FacebookSiteName siteName={ props.siteName } />
+		</Fragment>
 	);
 };
 
 FacebookPreview.propTypes = {
 	siteName: PropTypes.string.isRequired,
+	src: PropTypes.string.isRequired,
```
afercia: `src` required: see above

composites/Plugin/SocialPreviews/Facebook/helpers/determineFacebookImageProperties.js
```diff
@@ -0,0 +1,165 @@
...
+		img.onload = () => {
+			resolve( {
+				width: img.width,
+				height: img.height,
+			} );
+		};
+
+		img.onerror = reject;
+
+		img.src = src;
```
afercia: Promises are not my _forte_, but I'm not sure what `img.src = src;` is meant to do.
Also, on the "faulty image" example, I get an `Uncaught (in promise)` error in the console.

composites/Plugin/SocialPreviews/Facebook/helpers/determineFacebookImageProperties.js
```diff
@@ -0,0 +1,165 @@
...
+		return {
+			width: originalDimensions.width,
+			height: originalDimensions.height,
+		};
+	}
+
+	/*
+	 If the image is a square, just use the squareWidth and squareHeight. We don't have to fear
+	 that the resulting image will be warped.
+	 */
```
afercia: In a multi-line comment, lines between the first and last ones should start with an asterisk and a space.

composites/Plugin/SocialPreviews/Facebook/components/FacebookImage.js
```diff
@@ -0,0 +1,154 @@
...
+	position: relative;
+`;
+
+const FacebookImageContainer = styled.div`
+	line-height: 0;
+	position: relative;
+	z-index: 1;
+	overflow: hidden;
+	display: block;
+	background-color: #fff;
```
afercia: Please always use colors from the palette variables.

composites/Plugin/SocialPreviews/Facebook/components/FacebookImage.js
```diff
@@ -0,0 +1,154 @@
...
+import {
+	determineFacebookImageProperties,
+	LANDSCAPE_HEIGHT,
+	LANDSCAPE_WIDTH,
+	PORTRAIT_HEIGHT,
+	PORTRAIT_WIDTH,
+	SQUARE_HEIGHT,
+	SQUARE_WIDTH,
+} from "../helpers/determineFacebookImageProperties";
+import colors from "../../../../../style-guide/colors.json";
```
afercia: generally, we should try to keep external and internal dependencies separated in two blocks with comments.

afercia:
I've added a default, empty, image `alt` attribute. Without it, screen readers will try to read _something_ from the image, often the image file name:

screenshot 2019-02-11 at 11 14 48

We should then revisit this when the implementation will be complete. Passing the alt attribute set by users will allow to set the `og:image:alt`, see https://github.com/Yoast/yoast-components/issues/812#issuecomment-456851589. Also, in an authoring context, users need to know which image is set: when the image has no original alt attribute, we should use something like `alt="The current image has no alternative text. The file name is: %s"` only in the preview (note: this is what WordPress does in the image widget). Not sure if we can get the file name though.

Also:
- made two strings translatable and addressed a few minor things.
- changed `FacebookSiteNameWrapper` to a paragraph to be slightly more semantic: margins can be styled in the plugin

**Important considerations:**

1
I'm not sure I understand if there's the need to handle the initial state, when users haven't set an image yet. I see the image `src` is required and the initial state is `loading`. Maybe I'm missing some details.

2
Re: the layout: it's important to be aware the image is a flex item within a flex container, where the default value for `align-items` is `stretch`: this means images will be stretched by their height and this might affect the positioning / centering.

3
In Internet Explorer 11, the portrait image is not centered: I've tried to understand why, with no luck.

screenshot 182

afercia:
Maybe I've found a way to make Internet Explorer 11 happy. Will try it, then move back to Needs Action for feedback on the pending points.

afercia:
Too many flexbox bug in IE11 to center the images. Finally switched to absolute positioning and 
transform `translate()`, which fixes the issue.
- the images have now fixed width and height as calculated based on their ratio
- this means they can't be made responsive: they weren't even before these changes and I guess they shouldn't
- depending on the final implementation, there might be the need to update the image properties when the image is updated. At the moment, they're set only on `componentDidMount`

Test with images that are "very landscape" and "very portrait", for example:
```
http://via.placeholder.com/1200x300
http://via.placeholder.com/300x800
```

When changing images, refresh the page as the image props are set only on mount.

Pending feedback on the initial state (see previous comment).

afercia:
Pending feedback on:
- the initial state (see previous comment)
- not sure what `img.src = src;` in the promise is meant to do
- on the "faulty image" example, I've got an `Uncaught (in promise)` error in the console but couldn't reproduce consistently

IreneStr:
@afercia Thanks for your additions!

Some findings (in Chrome) based on your changes:
- The example portrait image is not centered horizontally. The `transform: translate( -50%,-50% );` doesn't do anything 
schermafbeelding 2019-02-12 om 10 08 18

- I recall you discussing an alt image input field in the UX channel/with the architects. Did you reach a coonclusion about whether or not we'd want such a field?

- The following will be addressed in a separate issue (which hasn't been created yet):
> I'm not sure I understand if there's the need to handle the initial state, when users haven't set an image yet. I see the image `src` is required and the initial state is `loading`. Maybe I'm missing some details.

- We'll handle the following as well in that issue:
> * depending on the final implementation, there might be the need to update the image properties when the image is updated. At the moment, they're set only on `componentDidMount`

- I've had the following error before I implemented this return in this catch: https://github.com/Yoast/yoast-components/pull/818/files#diff-cf84bc61f566d30a40d4c9261fb0306cR98
> * on the "faulty image" example, I've got an `Uncaught (in promise)` error in the console but couldn't reproduce consistently

afercia:
> The example portrait image is not centered horizontally. The `transform: translate( -50%,-50% );` doesn't do anything

@IreneStr please try to refresh the page. It does center the image in all browsers for me.

> I recall you discussing an alt image input field in the UX channel/with the architects. Did you reach a coonclusion about whether or not we'd want such a field?

Haven't. Anyways, we should make sure the final implementation renders an `og:image:alt` meta in the markup in the front-end. It would be nice to have it also in the preview.

IreneStr:
@afercia Refreshing, restarting the server, and clearing the cache all don't work for me, both in Chrome as well as Firefox 😕

afercia:
@IreneStr Looking into it.

afercia:
@IreneStr maybe not having a wrapper that centers the examples in the page doesn't help. The content is aligned to the left edge of the screen so noticing the image is actually centered is difficult. This is what I see when I add a wrapper to center the content:

screenshot 2019-02-12 at 15 48 18

Note: the CSS that first uses `position: absolute; top: 50%; left: 50%;` and then `translate()` is a standard technique to center content.

IreneStr:
The image is centered now 👍 
I think a final acceptance test by someone who didn't write a large part of the code would be in place, so I'm moving it to `needs-acceptance`.

Xyfi:
Acceptance 👍

